PR #28, opened 2017-09-19
[NARS] [estess] Fix test failures in T63000A due to V63001A code changes

nars1:
Test:
------
E_ALL run many times to test the fixes

README
-------
Below are general changes

* This is a project that updates T63000A (which is a copy of the test system that is in sync with GT.M V63000A)
  so it runs fine with GT.M V63001A (and the various code changes it comes with). Lots of test failures were fixed.
  Most of the test changes were to fix the reference files so they conform to the latest code changes.
  For example,
	a) New DBFILECREATED message prefix shows up whenever a database file is created.
	b) MUPIP INTEG output changed (due to block-split logic code changes in V63001A).
	c) The default GDS block size changed from 1K to 4K. This caused the same data to be held by fewer # of
		GDS blocks than before. This in turn affected the record number in MUPIP LOAD -FORMAT=BIN,
		MUPIP BACKUP etc. where each record is a block.
	d) The default jnl buffer size changed from 2308 to 2312. This also caused a new JNLBUFFDBUPD or
		JNLBUFFREGUPD message in some cases.
	e) The DSE DUMP -FILE output has a few more fields (asyncio, online freeze etc.).
	f) The GDE SHOW output has more columns and rework of previous column layout.
	g) Names of error messages issued in certain scenarios have changed. For example
		1) GVSUBSERR -> NOCANONICNAME in gde/outref/errors.txt
		2) NOREGION -> CLISTRTOOLONG in longname/outref/lock.txt
		3) JNLINVALID -> JNLSWITCHFAIL in mupjnl/u_inref/C9C04001977.csh
		4) MAXNRSUBSCRIPTS -> GVNAKED in v234/outref/v234.txt
		5) DBMISALIGN -> DBTOTBLK in v54002/outref/C9K11003341.txt
		6) REQROLLBACK message now prints MUPIP JOURNAL -ROLLBACK -NOONLINE (instead of MUPIP JOURNAL ROLLBACK)
	h) MUPIP SET now prints the database access method (MM or BG).
	i) MUPIP JOURNAL -SHOW has few new lines (Last EOF Written etc.)
	j) New REGFILENOTFOUND message shows up whenever a db file is not found
  The above applies to various tests.

* In some cases, the test scripts had to be fixed to account for the new behavior.
  For example, tests that relied on a 1K block size default now had their dbcreate.csh invocation changed to pass
	-block_size=1024 explicitly since the new default is 4K. This applies to various tests.

* Additionally, the block-split logic code changes caused the layout of the records to change.
  For example, the dse test relies on the positioning of records at certain blocks and offsets and that had to
	be fixed by finding where the same data is now placed and using that new block # and offset.
  This applies to various tests.

Below are specific changes that require a bit more explanation. Any change that is not explained below is
	a fix to the reference file to match the new code behavior.

* com/genmsgs.m : Handle messages with !AD!AD usage where the two !ADs are not separated by a space.
	ZGBLDIRACC was the only exception previously whereas ASYNCIONOMM is the new one so rework the
	code to avoid hardcoding of these exceptions. In these cases, make sure each !AD gets written without
	a newline in between but once all !ADs are written, all those pending newlines are written. So in
	the case of !AD!AD (ASYNCIONOMM), 2 newlines are written whereas in the case of !AD!AD!AD (ZGBLDIRACC),
	3 newlines are written.

* com/gtmtest.csh : sr_unix/check_trigger_support.csh no longer exists in V63001A. So assume triggers are always
	supported. This meant removing gtm_platform_triggers_support and PLATFORM_TRIGGER and PLATFORM_NOTRIGGER
	usages (in com/outref.awk, com/process.awk, longstr/outref/max_strlen.txt) as well.

* com/lotsvar.m : Ensure global names of the form ^%Y... are never generated as they are invalid from V63001A
	onwards (reserved for statsdb usages).

* collation/inref/ylct.m, errors/inref/erransi.m, pattern/inref/t.m, v62001/inref/gtm8174.m
	The compiler now evaluates literal expressions and therefore errors out at compile time whereas previously
	it did not. The M program was fixed to move the literal expression into an XECUTE to avoid the compiler error.

* increment/inref/incrgv4.m : Remove ZTSTART/ZTCOMMIT usages as that causes assert failures occasionally.
	> %GTM-F-ASSERT, Assert failed in sr_port/jnl_write.c line 501 for expression
		((!IS_TP(rectype) && !IS_ZTP(rectype)) || (0 != ((struct_jrec_upd *)jfb->buff)->update_num))

* mupip/u_inref/mu_set.csh : MUPIP SET -FILE -A conflicts with new -ASYNCIO qualifier so fully spell out -ACCESS_METHOD

* online_rollback/inref/nontp.m : Fix test since updates to non-replicated regions do not result in DBROLLEDBACK error
	in V63001A.

* read_only/u_inref/tstmrundown.csh AND read_only/u_inref/tstrundown.csh
	ftok semaphore is leftover in V63001A in case of a DBRDONLY error from mupip rundown.
	Therefore fix test to give read-write permissions and redo rundown before proceeding to avoid test
	failures when it later checks leftover ipcs.

* repl_misc_qual/u_inref/repl_misc_qual.csh : In V63001A, MUPIP REPLICATE CHANGELOG returns status of 0 if no log
	change happens. Take this opportunity to also reduce test runtime.

* spanning_nodes/u_inref/dse.csh : Default block size is now 4K so increase the 3000 record size (used previously
	to guarantee a spanning node with a block size of 1K) to be more than 4K.

* v43001d/u_inref/D9C08002195.csh : sort the output to account for ftok-order of region output (which is
	non-deterministic as it is based on the inode of the .dat file)

* v53003/inref/c003044.m, v62002/outref/gtm8087.txt : Account for new BTD statistic.

* v54002/inref/C9B04001673a.m : Fix M program so the V63001A behavior is considered as a PASS

* v54002/inref/D9K05002773.m AND v54002/u_inref/D9K05002773.csh
	v54002/D9K05002773 test relies on index block fill factor of 90% which in turn relies on 1K block size
	so use 1K blocksize explicitly in dbcreate.csh invocation. Remove test_specific_gde since that causes
	dbcreate.csh to not honor input

* v54003/outref/C9K11003340.txt : There is one less timer now (heartbeat timer has been removed in V63001A).

* v54003/u_inref/resync.csh : The trace buffer sizes have changed from 1Mb to 4Mb so account for that in the test.
	Or else fetchresync rollback will show up as having taken a lot more memory than resync rollback.

* v60000/inref/antifreezeutil.m : With V63001A, the result of a sequence of updates is that a reorg needs to do
	nothing. But this test relies on reorg actually modifying one block (and in turn noticing the instance freeze)
	so enhance the test to do some kills thereby making reorg do real work.

* v60000/u_inref/gtm4525c.csh : DBDANGER does not show up. Most likely due to code changes in V63001A. So remove that
	portion of the test.

* v60002/inref/peekaboo.m : Remove section that tested lower-case region names as they now imply a statsdb region
	in V63001A.

* v60002/outref/peekaboo.txt : Default block size changed from 1K to 4K

* v60003/u_inref/gtm7528.csh : Remove gtm_white_box_test_case_count=1 scenario as that is no longer relevant with
	V63001A rework of db_init() (see script for detailed comments). Remove search of REGOPENRETRY message as
	it no longer exists. Wait for backgrounded dse processes to finish before exiting from test.

* v62000/u_inref/gtm8086.csh : Processes now fail with an additional JNLSWITCHFAIL message in V63001A so take
	that into account in the test script.

* v62002/u_inref/gtm6638.csh : GDE SHOW output has changed so rework test to search for epoch-taper field.

* v63000/inref/gtm8296.m : PEEKBYNAME("gd_region.jnl_state") does not open the database in V63001A so enhance test
	script to verify this as well as open the database explicitly before redoing this and continuing with the
	test's original purpose.
* mprof/outref/general.txt : Fix mprof test failure due to <quit:0> no longer being traced by M-profiling in V63001A


nars1:
No need anymore for this pull request since a similar pull request has been merged into the T63001A branch of YottaDB/YottaDBtest.

PR #50, opened 2017-11-10
[NARS1] [estess] Fix v62001/gtm8206 test failure due to code now runn…

nars1:
…ing lot faster than before

The test did the below (in an expect script).

	for i=1:1:2000000 set a(i)=i
	merge b=a

And sends a Ctrl-C 2 seconds after the merge command has started. That usually takes 8 seconds to run even on a fast server and so the Ctrl-C always happens and b(2000000) is never set when the test checks it later. But with the latest stpg_sort() performance enhancements (see https://github.com/YottaDB/YottaDB/issues/85), the merge finished in 0.7 seconds and so the Ctrl-C was a no-op. And disturbed the timing that the test was expecting resulting in a failure.

The test is now fixed by changing the "merge b=a" to a "for i=1:1  merge b(i)=a" so we are guaranteed it is an infinite loop however fast a machine is. And after 2 seconds the test will interrupt this process and check if a GTM-I-CTRLC message shows up (like it used to). But the check that b(2000000) is indeed undefined is now removed. That is an indirect verification of the fact that the for loop did not run to completion due to the Ctrl-C and not necessary in my opinion since we are about to issue a "halt" soon after and if the loop indeed did not get interrupted, the halt would not get control and the test would not finish.

### DIFF
--- a/v62001/outref/gtm8206.txt
+++ b/v62001/outref/gtm8206.txt
@@ -1,3 +1,2 @@
-# Running expect (output: expect.out)
-1
-1
+# Running expect (output: expect.out) : Expecting a GTM-I-CTRLC message
+%GTM-I-CTRLC, CTRL_C encountered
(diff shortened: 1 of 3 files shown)


shabiel:
That's amazing!

nars1:
Thanks Sam.

PR #71, opened 2017-12-26
Fix v63000/gtm8394 subtest failure due to pthread_cancel fatal/abort message from libc; Ensure this message goes to stderr even if test was started from a terminal

nars1:


v63000/u_inref/gtm8394.csh
```diff
@@ -14,6 +17,21 @@
...
+#
+#	libgcc_s.so.1 must be installed for pthread_cancel to work
+#
+# This message is issued on the terminal where this test was issued from even if the test stdout/stderr have been
+# redirected to a file. This is because the above message is treated as a fatal libc error and those go through
+# a special function that tries to open the terminal (if available) to log the error. The following env var
+# overrides that logic and continue to log this error in stderr.
+#
+setenv LIBC_FATAL_STDERR_ 1
+set gccabort = "libgcc_s.so.1 must be installed for pthread_cancel to work"
```
estess: My concern here is grep-ping for a specific error that includes a specific version of a libgcc library. At any time this library level could be different - and could potentially even be different on different distros. The message could change too but that's less likely than the library version. Perhaps wildcard the version id?

PR #91, opened 2018-02-20
[#83] New r120/readcmdrecallhist subtest to test M READ Editing in terminal has history of 99 lines (same as direct mode)

nars1:


nars1:
expect also supports associative arrays. How about that :). It took a while but it was fun learning to program with expect. All these interactive tasks are easily automated.

PR #96, opened 2018-03-02
Work around an lsof 4.90 bug (where lsof of a specific file dumped extra output) by grepping for the specific file again

nars1:


v54000/u_inref/D9J07002728.csh
```diff
@@ -19,7 +31,8 @@ echo "# Wait until object file is created"
 $gtm_tst/com/wait_for_log.csh -log badcomp.o
 echo "# Output from lsof means badcomp.o is open (and probably locked).  We expect no output."
 ($lsof $tst_working_dir/badcomp.o > lsof.outx) >& lsof.err
-cat lsof.outx
+# lsof 4.90 had a bug where it listed extraneous output so filter that out by a grep of badcomp.o
```
estess: Just a nit here - I'd imagine lsof 4.90 *still* has the bug and will for all time. The fix is in another version so the past-tense here looks a bit strange. But it does convey it's meaning so fix if you care.

nars1: Sorry did not realize this tense mismatch. Since this is a cosmetic change I will not go through another review for this. Thanks.

PR #101, opened 2018-03-17
Test for ydb_maxtptime and ydb_dbglvl

nars1:


r120/outref/ydb_dbglvl.txt
```diff
@@ -0,0 +1,10 @@
+# Test for ydb_dbglvl env var
+# Test that gtmdbglvl=2 (print usage statistics at process end) controls ydbDebugLevel if ydb_dbglvl is undefined
```
estess: Why use ydbDebugLevel here? The name of the internal variable is still gtmDebutLevel - likely for a long time to come.

r120/u_inref/ydb_dbglvl.csh
```diff
@@ -0,0 +1,33 @@
...
+#								#
+#################################################################
+#
+echo "# Test for ydb_dbglvl env var"
+
+#
+echo '# Test that gtmdbglvl=2 (print usage statistics at process end) controls ydbDebugLevel if ydb_dbglvl is undefined'
+unsetenv ydb_dbglvl
+setenv gtmdbglvl 2
+$gtm_exe/mumps -run %XCMD 'write "ydbDebugLevel=",ydbDebugLevel,!' |& $grep "small storage performance"
```
estess: Why use a code fragment that generates an error here (ydbDebugLevel is not defined) instead of just doing say "set a=1+2" for example.

r120/outref/ydb_dbglvl.txt
nars1: To address this, I have made the code changes to say it is ydbDebugLevel in the simpleAPI branch (#59 which also houses the code changes for #115).

r120/u_inref/ydb_dbglvl.csh
nars1: I did not even realize the error until you mentioned. But thinking about it, since I am grepping for a specific string whose output is dependent only on ydb_dbglvl setting and not on the actual M line executed, I hope you are okay with any M line. Let me know if not and I will fix it.

r120/outref/ydb_dbglvl.txt
estess: I'm still missing something as my confusion runs rampant. The gtmDebugLevel variable is a C global variable not accessible from M code so unless you changed it yesterday and I don't see it yet, then it is still gtmDebugLevel even in the simpleAPI version so has not been changed, nor is there any reason to change the global variable name.

nars1: It was a typo in a comment line in the test script. And so it did not matter whether it was ydbDebugLevel or gtmDebugLevel. Now that I have made the ydbDebugLevel change, what was a typo is no longer one. Hope this clears your confusion.

PR #121, opened 2018-04-26
Changed gtm8791.exp ...

JakeYottaDB:
In order to push a timeout message and exit upon a timed out expect statement. The syntax of the send statement has also been changed so that a send statement will only be executed upon a successful expect statement.

### DIFF
--- a/v63004/u_inref/gtm8791.exp
+++ b/v63004/u_inref/gtm8791.exp
@@ -13,20 +13,26 @@
 set timeout 60
 spawn /usr/local/bin/tcsh -f
 
+expect_after {
+	timeout { timeout_procedure }
+}
+
+proc timeout_procedure { } {
+	puts "timeout occurred"
+	exit -1
+}
+
+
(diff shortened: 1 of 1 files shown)


v63004/u_inref/gtm8791.exp
```diff
@@ -13,20 +13,26 @@
...
+	exit -1
+}
+
+
 send -- "\$gtm_dist/lke\r"
 
-expect -exact "LKE>"
 # Send <Ctrl-Z>
-send -- "\x1A\r"
+expect -exact "LKE>" {send -- "\x1A\r"}
```
nars1: Now that we will exit the expect script the moment a timeout happens, do we need the change in this line? Seems like we can have expect commands and send commands in separate lines like before.

JakeYottaDB: If thats preferable I'll go ahead and change the send lines back to the former syntax.

PR #147, opened 2018-05-08
 New v63004/gtm8909 subtest (tests GTM-8909 in V63004) 

JakeYottaDB:
The help facility for the utility programs MUPIP, DSE and LKE ignores control-C; previously if a user pressed <CTRL-C> while using help, the facility could exit with a harmless, but inappropriate, ENO256 error.


v63004/u_inref/gtm8909.exp
```diff
@@ -0,0 +1,73 @@
...
+expect -exact ">"
+puts "# Start LKE help facility"
+send -- "\$gtm_dist/lke\r"
+expect -exact "LKE>"
+send -- "help\r"
+puts "# Expect Help Facility Prompt"
+expect  "Topic?"
+puts "#<Ctrl-C>"
+send -- "\x03\r"
+send -- "\r"
```
nars1: What is the need for this \r when we already have a \r in the previous line.

v63004/u_inref/gtm8909.exp
```diff
@@ -0,0 +1,73 @@
...
+puts "# Exit MUPIP>"
+expect -exact "MUPIP>"
+send -- "exit\r"
+
+puts "# Expect the shell prompt"
+expect -exact ">"
+
+interact
```
nars1: What does this interact command do?

v63004/u_inref/gtm8909.exp
```diff
@@ -0,0 +1,73 @@
...
+puts "#<Ctrl-C>"
+send -- "\x03\r"
+send -- "\r"
+puts "# Exit LKE>"
+expect -exact "LKE>"
+send -- "exit\r"
+
+puts "# Expect the shell prompt"
+expect -exact ">"
+puts "# Start DSE help facility"
```
nars1: The script flow is the same for LKE, DSE and MUPIP. It would be nice (not necessary) to fold this into a loop (for or while etc.) so we avoid code duplication. I leave the choice to you.

v63004/u_inref/gtm8909.exp
```diff
@@ -0,0 +1,73 @@
...
+	exit -1
+}
+
+puts "# Expect the shell prompt"
+expect -exact ">"
+puts "# Start LKE help facility"
+send -- "\$gtm_dist/lke\r"
+expect -exact "LKE>"
+send -- "help\r"
+puts "# Expect Help Facility Prompt"
```
nars1: This puts can be removed. It does not add much value particularly when one goes through the reference file where it moves the "help" comand that the user enters into the line after the LKE> prompt so makes it harder to understand the flow.

v63004/u_inref/gtm8909.exp
```diff
@@ -22,52 +24,54 @@ proc timeout_procedure { } {
...
-send -- "\r"
-puts "# Exit MUPIP>"
-expect -exact "MUPIP>"
-send -- "exit\r"
 
-puts "# Expect the shell prompt"
-expect -exact ">"
-
-interact
+#puts "# Expect the shell prompt"
```
nars1: Can the below commented out section be removed.

v63004/u_inref/gtm8909.exp
```diff
@@ -22,52 +24,54 @@ proc timeout_procedure { } {
...
-send -- "help\r"
-puts "# Expect Help Facility Prompt"
-expect 	"Topic?"
-puts "# <Ctrl-C>"
-send -- "\x03\r"
-send -- "\r"
-puts "# Exit DSE>"
-expect -exact "DSE>"
-send -- "exit\r"
+foreach prompt $promptList {
```
nars1: Thank you. The foreach loop looks much better.

PR #148, opened 2018-05-09
gtm8909 fix

JakeYottaDB:
Set terminal columns explicitly in gtm8909.exp, Filter out blank lines in gtm8909.csh

### DIFF
--- a/v63004/u_inref/gtm8909.csh
+++ b/v63004/u_inref/gtm8909.csh
@@ -23,7 +23,6 @@ mv expect.out expect.outx	# move .out to .outx to avoid -E- from being caught by
 perl $gtm_tst/com/expectsanitize.pl expect.outx > expect_sanitized.outx
 # The output is variable on slow vs fast systems and so filter out just the essential part of it to keep it deterministic.
 
-
-cat expect_sanitized.outx
+$grep -P "\S" expect_sanitized.outx
 
 $gtm_tst/com/dbcheck.csh >> db_log.txt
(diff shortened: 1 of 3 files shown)


v63004/u_inref/gtm8909.csh
```diff
@@ -23,7 +23,6 @@ mv expect.out expect.outx	# move .out to .outx to avoid -E- from being caught by
 perl $gtm_tst/com/expectsanitize.pl expect.outx > expect_sanitized.outx
 # The output is variable on slow vs fast systems and so filter out just the essential part of it to keep it deterministic.
 
-
-cat expect_sanitized.outx
+$grep -P "\S" expect_sanitized.outx
```
nars1: grep -P man pages say this is highly experimental (that said, knowing grep can use perl type regexp searching is a good find). So use instead $grep -v '^$'. This is a simple way to filter out empty lines. Also add a comment before this line indicating we have seen extra empty lines in rare cases of test runs on slower systems (and don't have a precise understanding of the cause) and since they are not essential, we are filtering those out.

v63004/u_inref/gtm8909.exp
```diff
@@ -24,10 +24,13 @@ proc timeout_procedure { } {
 	exit -1
 }
 
+send -- "stty cols 130\r"
```
nars1: Add a comment before this line as to WHY this is done (i.e. something like what we have in r120/u_inref/ctrlchandler.exp).

v63004/u_inref/gtm8909.csh
JakeYottaDB: grep -v '^$' has now been implemented with an appropriate comment

v63004/u_inref/gtm8909.exp
JakeYottaDB: An explanatory comment has been added

PR #151, opened 2018-05-14
New v63004/gtm5730 subtest (tests GTM-5730 in V63004)

JakeYottaDB:
The Update Process logs record types with their corresponding type description; previously it only logged the integer type value. Note that it is possible, the change could disrupt code that parses the modified results.


v63004/inref/gtm5730.m
```diff
@@ -0,0 +1,21 @@
...
+;	of its copyright holder(s), and is made available	;
+;	under a license.  If you do not know the terms of	;
+;	the license, please stop and do not read further.	;
+;								;
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;
+gtm5730()
+	 TSTART
+	 FOR i=1:1:2000 set ^x(i)=i ;
+	 FOR i=1:1:2000 W ^x(i);
```
nars1: What purpose does this line (which dumps a lot of data) serve?

v63004/inref/gtm5730.m
```diff
@@ -0,0 +1,21 @@
...
+;	This source code contains the intellectual property	;
+;	of its copyright holder(s), and is made available	;
+;	under a license.  If you do not know the terms of	;
+;	the license, please stop and do not read further.	;
+;								;
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;
+gtm5730()
+	 TSTART
+	 FOR i=1:1:2000 set ^x(i)=i ;
```
nars1: Since we do the 2000 SETs inside a TSTART/TCOMMIT, we are doing just ONE transaction/sequence-number. It is better to move the FOR loop outside the TSTART/TCOMMIT logic. And do 1000 of TSTART/TCOMMIT and 1000 without TSTART/TCOMMIT. That way we can expect to see TCOM as well as SET records in the reference file. In addition, it would be better to see KILL record too in the reference file.

v63004/inref/gtm5730.m
JakeYottaDB: I've removed the writing forloop

v63004/inref/gtm5730.m
JakeYottaDB: I've split the .m file into 2 forloops for 1000 individual TSTART/TCOMMIT transactions and 1000 regular transactions respectively. ^x is killed at the end of the m program to generate the kill type record.

PR #165, opened 2018-05-21
New release for v63004/gtm8891. Tests the following release note: A <…

VinayYottaDB:
…side-effect-expression><pure-Boolean-operator>$SELECT(0:side-effect-expression)) sequence produces a SELECTFALSE run-time error; a regression introduced with the literal optimizations in V6.3-002 caused this combination to produce a SIG-11 (segmentation violation) at compilation. (GTM-8891)

### DIFF
--- a/v63004/inref/gtm8891.m
+++ b/v63004/inref/gtm8891.m
@@ -1,0 +1,19 @@
+#
+#################################################################
+#								#
+# Copyright (c) 2018 YottaDB LLC. and/or its subsidiaries.	#
+# all rights reserved.						#
+#								#
+#	This source code contains the intellectual property	#
+#	of its copyright holder(s), and is made available	#
+#	under a license. If you do not know the terms of	#
+#	the license, please stop and do not read further.	#
+#								#
+#################################################################
+#
+
+
+gtm8891()
+	w $incr(x)!$select(0:$incr(x))
+	quit
+
--- a/v63004/inref/gtm8891.m
+++ b/v63004/inref/gtm8891.m
@@ -1,19 +1,17 @@
-#
-#################################################################
-#								#
-# Copyright (c) 2018 YottaDB LLC. and/or its subsidiaries.	#
-# all rights reserved.						#
-#								#
-#	This source code contains the intellectual property	#
-#	of its copyright holder(s), and is made available	#
-#	under a license. If you do not know the terms of	#
-#	the license, please stop and do not read further.	#
-#								#
-#################################################################
-#
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;								;
+; Copyright (c) 2018 YottaDB LLC. and/or its subsidiaries.	;
+; all rights reserved.						;
+;								;
+;	This source code contains the intellectual property	;
+;	of its copyright holder(s), and is made available	;
+;	under a license. If you do not know the terms of	;
+;	the license, please stop and do not read further.	;
+;								;
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
 
 gtm8891()
-	w $incr(x)!$select(0:$incr(x))
+	write $incr(x)!$select(0:$incr(x))
 	quit
 
--- a/v63004/instream.csh
+++ b/v63004/instream.csh
@@ -2,25 +2,26 @@
 #################################################################
 #								#
 # Copyright (c) 2018 YottaDB LLC. and/or its subsidiaries.	#
-# All rights reserved.						#
+# all rights reserved.						#
 #								#
 #	This source code contains the intellectual property	#
 #	of its copyright holder(s), and is made available	#
-#	under a license.  If you do not know the terms of	#
+#	under a license. If you do not know the terms of	#
 #	the license, please stop and do not read further.	#
 #								#
 #################################################################
 #
 #-----------------------------------------------------------------------------------------------------------------------------
-# List of subtests of the form "subtestname [author] description"
+# list of subtests of the form "subtestname [author] description"
 #-----------------------------------------------------------------------------------------------------------------------------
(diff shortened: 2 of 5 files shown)


v63004/inref/gtm8891.m
```diff
@@ -0,0 +1,19 @@
+#
```
nars1: Remove line 1. Our convention is to have the ####### line as the first line in the file.

v63004/instream.csh
```diff
@@ -2,33 +2,34 @@
 #################################################################
 #								#
 # Copyright (c) 2018 YottaDB LLC. and/or its subsidiaries.	#
-# All rights reserved.						#
+# all rights reserved.						#
```
nars1: This seems to be an unintended change. Please undo.

v63004/instream.csh
```diff
@@ -2,33 +2,34 @@
...
 #								#
 #	This source code contains the intellectual property	#
 #	of its copyright holder(s), and is made available	#
-#	under a license.  If you do not know the terms of	#
+#	under a license. If you do not know the terms of	#
```
nars1: A space before the If was removed unintentionally. Please undo.

v63004/instream.csh
```diff
@@ -2,33 +2,34 @@
...
 #	the license, please stop and do not read further.	#
 #								#
 #################################################################
 #
 #-----------------------------------------------------------------------------------------------------------------------------
-# List of subtests of the form "subtestname [author] description"
+# list of subtests of the form "subtestname [author] description"
```
nars1: Same issue with "list"

v63004/instream.csh
```diff
@@ -2,33 +2,34 @@
...
 #-----------------------------------------------------------------------------------------------------------------------------
-# gtm8909	    [jake]  Tests that <ctrl-c> within the help facility no longer leads to EN0256 error upon exit
-# gtm8860	    [jake]  Tests that journal extract removes additional / from journal and output file paths
-# gtm8791	    [jake]  Tests that <ctrl-z> no longer causes segmentation violation
-# gtm8699	    [jake]  Tests that $VIEW("STATSHARE",<region>) returns 1 if the process is sharing DB stats and 0 otherwise
-# gtm8202	    [jake]  Tests the functionality of the -SEQNO qualifier for the mupip journal -extract command
-# gtm5730	    [jake]  Tests that the update process now logs record types with a corresponding, non-numerical, description
-# gtm1041	    [jake]  Tests the that env variable gtm_mstack_size sets the size of the M stack as expected
+# gtm8909	    [jake]  tests that <ctrl-c> within the help facility no longer leads to en0256 error upon exit
```
nars1: I see a huge diff block here. I was expecting just the one line corresponding to the newly added subtest. What was changed in the other lines?

v63004/u_inref/gtm8891.csh
```diff
@@ -0,0 +1,17 @@
+
```
nars1: Remove line 1. Same as a prior comment.

v63004/u_inref/gtm8891.csh
```diff
@@ -0,0 +1,17 @@
...
+#	of its copyright holder(s), and is made available	#
+#	under a license. If you do not know the terms of	#
+#	the license, please stop and do not read further.	#
+#								#
+#################################################################
+#
+
+echo "#Running gtm8891 test"
+$gtm_dist/mumps -run gtm8891
+$gtm_dist/mumps -run %XCMD 'halt'
```
nars1: Line 17 is not necessary. The mumps -run done in line 16 returns back to the shell so no need to start a new mumps process which only does a halt.

v63004/inref/gtm8891.m
```diff
@@ -0,0 +1,19 @@
...
+#	of its copyright holder(s), and is made available	#
+#	under a license. If you do not know the terms of	#
+#	the license, please stop and do not read further.	#
+#								#
+#################################################################
+#
+
+
+gtm8891()
+	w $incr(x)!$select(0:$incr(x))
```
nars1: w -> write. It is a convention to include non-abbreviated commands in M code.

whitten:
I don't see a problem with $S(0:<expr>) generating an error. In fact, the M standard pretty much requires that it happen.   My problem is that the error is supposed to be "M4" as it says here:
http://71.174.62.16/Demo/AnnoStd?Frame=Main&Page=a107107&Edition=1995

7.1.5.18 $Select

$S[ELECT] ( L | tvexpr : expr | )

This form returns the value of the leftmost expr whose corresponding tvexpr is true. The process of evaluation consists of evaluating the tvexprs, one at a time in left-to-right order, until the first one is found whose value is true. The expr corresponding to this tvexpr (and no other) is evaluated and this value is made the value of $Select. An error condition occurs, with ecode="M4", if all tvexprs are false. Since only one expr is evaluated at any invocation of $Select, that is the only expr which must have a defined value.

If this situation produces  a SELECTFALSE run-time error, I hope the $ECODE value will be "M4"

v63004/instream.csh
VinayYottaDB: The tests are listed in reverse numeric order. Since 8891 is second from the top, all the lines below it are displaced by 1 line which causes these changes

nars1:
@whitten : Yes $ECODE will be M4

v63004/instream.csh
```diff
@@ -14,21 +14,22 @@
 #-----------------------------------------------------------------------------------------------------------------------------
 # List of subtests of the form "subtestname [author] description"
 #-----------------------------------------------------------------------------------------------------------------------------
-# gtm8909	    [jake]  Tests that <ctrl-c> within the help facility no longer leads to EN0256 error upon exit
```
nars1: There is now a huge block of prior commits that have been removed inadvertently. Please re-add them.

v63004/instream.csh
nars1: The displacement of lines is not the issue. There have been a lot of upper case letters changed to lower case. That is the reason for the huge block diff and needs to be reverted to their original case.

v63004/instream.csh
nars1: I realize now the comment lines have not been removed. They have been moved to a different place but with some upper case letters changed to lower case.  Never mind this comment. The other comment still needs to be addressed.

PR #167, opened 2018-05-22
New v63004/gtm8894 subtest (tests GTM8894 in V63004)

VinayYottaDB:

The $ZRELDATE Intrinsic Special Variable provides the UTC date / time of the build GT.M build in the form YYYYMMDD 24:60 (using $ZDATE() notation). While $ZVERSION is probably a better identifier for most uses, $ZRELDATE may be a helpful alternative for those testing pre-release builds.

### DIFF
--- a/v63004/outref/gtm8894.txt
+++ b/v63004/outref/gtm8894.txt
@@ -1,3 +1,3 @@
 # Generating output for $zreldate
 # Verify output is in libyottadb##TEST_SHL_SUFFIX## and is correct format
-20180518 15:42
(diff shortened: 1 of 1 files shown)


v63004/outref/gtm8894.txt
```diff
@@ -1,3 +1,3 @@
 # Generating output for $zreldate
-# Verify output is in libyottadb##TEST_SHL_SUFFIX##
+# Verify output is in libyottadb##TEST_SHL_SUFFIX## and is correct format
 20180518 15:42
```
nars1: This timestamp still needs to be replaced by a TEST_AWK expression.

PR #168, opened 2018-05-22
New v63004/gtm8914 subtest (tests GTM-8699 in V63004)

JakeYottaDB:

$VIEW("GVSTATS",<region>), ZSHOW "G" and ZSHOW "T" 
return a question-mark (?) at the end of their output strings when the process does not have access to the current shared statistics; 
they did not do this previously. Note that it is possible, although unlikely, the flag character could disrupt code that parses the results of these facilities.


v63004/outref/gtm8914.txt
```diff
@@ -0,0 +1,17 @@
+# Enable sharing with gtm_statshare
+# Set gtm_statsdir to the root directory, which we dont have permissions to
+# Create a single region DB with gbl_dir mumps.gld and region DEFAULT
+# The DB, while set for sharing, should now be unable to share due to the invalid gtm_statsdir selection
+
+# Run gtm8914.m
+# End of line for each commands output: 
+$VIEW("GVSTATS","DEFAULT")
+0?
+ZSHOW
```
nars1: Why is ZSHOW "G" split into two lines? Same with ZSHOW "T".

v63004/u_inref/gtm8914.csh
```diff
@@ -24,16 +24,28 @@ echo "# The DB, while set for sharing, should now be unable to share due to the
...
-	if ("$cmd" != "") then
-		echo $cmd
-	else
-		echo "$line" | awk -F":" '{print $NF}'
-	endif
-
-end
+$ydb_dist/mumps -run gtm8914
 
+#echo '# End of line for each commands output: '
```
nars1: Can the big block of commented out lines be removed. Thanks.

PR #174, opened 2018-05-23
New v63004/gtm8643 subtest (tests GTM-8643 in V63004)

JakeYottaDB:

The listen queue depth specified with WRITE /LISTEN for a listening socket is limited only by the limit imposed by the OS; previously, GT.M enforced an artificial limit of 5. A SOCKET device retries connection attempts on local sockets that fail due to possible transient issues on the other end of the attempted connection, (for example an insufficiently large listen queue). Previously such attempts could sometimes appear to succeed, when they actually did not, leading to subsequent errors from READ and WRITE on the socket. (GTM-8643)


v63004/instream.csh
```diff
@@ -25,13 +25,14 @@
 # gtm1042	    [jake]  Tests the that env variable gtm_mstack_size sets the size of the M stack as expected
 # gtm8891	    [vinay] Tests that <side-effect-expression><pure-Boolean-operator>$SELECT(0:side-effect-expression)) sequence produces a SELECTFALSE runtime error
 # gtm8894	    [vinay] Tests that $zreldate outputs in the form YYYYMMDD 24:60
+# gtm8643	    [jake]  Tests that YDB no longer enforces queue depth limit of 5 for SOCKET devices. NOT TESTED YET (Confirms connection attempts fail upon transient issues)
```
JakeYottaDB: I've tried to specify in the instream.csh comment that the latter portion of the subtest hasn't been made yet. I figured I should make you aware of this, as you may find it unnecessary or undesirable.

v63004/u_inref/gtm8643.csh
```diff
@@ -0,0 +1,31 @@
...
+#								#
+#	This source code contains the intellectual property	#
+#	of its copyright holder(s), and is made available	#
+#	under a license.  If you do not know the terms of	#
+#	the license, please stop and do not read further.	#
+#								#
+#################################################################
+#
+
+#echo "# Enable sharing with gtm_statshare"
```
nars1: Please remove the big block of commented lines.

v63004/inref/gtm8643.m
```diff
@@ -0,0 +1,44 @@
...
+;
+gtm8643
+ 	set tcpdev="server$"_$j
+ 	set timeout=10
+ 	set tcpdev="server$"_$j
+ 	set portno=0
+
+ 	open tcpdev:(ZLISTEN=portno_":TCP":attach="server"):timeout:"SOCKET"
+
+ 	use $p	; principle device
```
nars1: principle -> principal. Needs to be fixed in a few other places too.

v63004/inref/gtm8643.m
```diff
@@ -0,0 +1,44 @@
...
+;								;
+;	This source code contains the intellectual property	;
+;	of its copyright holder(s), and is made available	;
+;	under a license.  If you do not know the terms of	;
+;	the license, please stop and do not read further.	;
+;								;
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;
+gtm8643
+ 	set tcpdev="server$"_$j
```
nars1: Lines 14 and 16 seem to be duplicates.

v63004/inref/gtm8643.m
```diff
@@ -0,0 +1,44 @@
...
+ 	set portno=0
+
+ 	open tcpdev:(ZLISTEN=portno_":TCP":attach="server"):timeout:"SOCKET"
+
+ 	use $p	; principle device
+	WRITE "WRITE /listen(4):",!
+ 	use tcpdev
+ 	WRITE /listen(4)
+ 	WRITE /wait(timeout)
```
nars1: Why do we need the WRITE /wait? Does that wait for timeout seconds i.e. 10 seconds before returning? If so, I would like to avoid that unnecessary step.

v63004/inref/gtm8643.m
```diff
@@ -0,0 +1,44 @@
...
+	WRITE !
+
+ 	use $p	; principle device
+	WRITE "WRITE /listen(5):",!
```
nars1: The 4, 5 and 6 second waits can all be folded into a loop to avoid code duplication.

v63004/instream.csh
nars1: Thanks. That helps. I will look into it to see if the rest is easily testable or not.

v63004/outref/gtm8643.txt
```diff
@@ -0,0 +1,4 @@
+# Run gtm8643.m to test queue depth limit
```
nars1: queue depth limit -> queue depth limit > 5 works fine

v63004/inref/gtm8643.m
```diff
@@ -11,33 +11,18 @@
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;
 gtm8643
- 	set tcpdev="server$"_$j
- 	set timeout=10
  	set tcpdev="server$"_$j
  	set portno=0
+	set timeout=10
```
nars1: How long does the test take to run? Does it run faster with a timeout of 1 (instead of 10 seconds)? If so, I would prefer 1. If not, then I am fine with the current changes.

JakeYottaDB: So it seems like the timeout variable is never used to wait for that amount of time, but is perhaps used to set the tcpdev up. The test ran in 13.66 seconds with the timeout=10 and 13.95 seconds with timeout=1. So it does not seem to have any effect on the run time.

v63004/instream.csh
nars1: I took a look at this and was not able to come up with a simple test for the retry issue the release note talks about. So please add a note that only the first sentence in the release note is tested.

v63004/inref/gtm8643.m
nars1: Then I am fine with the current changes.

PR #177, opened 2018-05-24
Test for v63003/gtm8804 (Tests GTM-8804 in V63003)

VinayYottaDB:

The ZSHOW "T" (where "T" can be case-insensitive) produces only the summary lines for "G" and "L" output; previously ZSHOW always showed the detail with the summary. (GTM-8804)


v63003/inref/gtm8804.m
```diff
@@ -0,0 +1,38 @@
...
+
+	write !,"output for zshow l",!
+	ZSHOW "L"
+
+	write !,"output for zshow t (summary of g and l)",!
+	new t
+	ZSHOW "T":t
+	;filter out variable settings
+	do out^zshowgfilter(.t,"CTN,DFL,DFS,JFL,JFS,JBB,JFB,JFW,JRL,JRP,JRE,JRI,JRO,CAT,CFE,CFS,CFT,CQS,CQT,CYS,CYT,BTD")
+	write t("G",0),!
```
nars1: Need to do a ZWRITE of t. That will dump the entire "t" variable. And is what will confirm that there is only a summary line for ZSHOW "T". Similarly, line 25 needs to do a ZWRITE g. That will confirm ZSHOW "G" dumps summary and detail lines.

v63003/inref/gtm8804.m
```diff
@@ -0,0 +1,38 @@
...
+;	under a license.  If you do not know the terms of	;
+;	the license, please stop and do not read further.	;
+;								;
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;
+; Tests that ZSHOW "T" produces summary lines for "G" and "L"
+;
+
+gtm8804
+	set ^x=1
```
nars1: Add a comment that this is done to ensure ZSHOW "G" has at least one non-zero statistic. Likewise, add a comment that line 18 is done to ensure ZSHOW "L" has at least one non-zero statistic.

PR #180, opened 2018-05-25
Test for v63003/gtm8617 (Tests GTM-8617 in V63003)

VinayYottaDB:

The MUPIP SET command supports the following qualifiers: -N[ULL_SUBSCRIPTS]={never, always, existing}, which controls whether GT.M accepts null subscripts for database keys. -[NO]STD[NULLCOLL], which determines whether GT.M will use standard MUMPS collation or GT.M collation for null-subscripted keys. Previously, this functionality was only available through GDE for database file creation, and DSE for existing database files. FIS strongly recommends avoiding the use of DSE when there is an alternative. (GTM-8617)

### DIFF
--- a/v63003/instream.csh
+++ b/v63003/instream.csh
@@ -19,7 +19,7 @@
 # gtm8186	    [vinay] Test DO, GOTO and ZGOTO can take offset without a label
 # gtm8804	    [vinay] Test zshow "t" produces only a summary of zshow "g" and "l"
 # gtm8832	    [vinay] Test string literal evaluating to >=1E47 produces a NUMOFLOW error
-# gtm8617	    [vinay] Tests functionality of STDNULLCOLL and NULL_SUBSCRIPTS
+# gtm8617	    [vinay] Tests MUPIP SET command of STDNULLCOLL and NULL_SUBSCRIPTS
 #-------------------------------------------------------------------------------------
 
 echo "v63003 test starts..."
--- a/v63003/instream.csh
+++ b/v63003/instream.csh
@@ -19,13 +19,14 @@
 # gtm8186	    [vinay] Test DO, GOTO and ZGOTO can take offset without a label
 # gtm8804	    [vinay] Test zshow "t" produces only a summary of zshow "g" and "l"
 # gtm8832	    [vinay] Test string literal evaluating to >=1E47 produces a NUMOFLOW error
+# gtm8617	    [vinay] Tests functionality of STDNULLCOLL and NULL_SUBSCRIPTS
 #-------------------------------------------------------------------------------------
 
 echo "v63003 test starts..."
 
 # List the subtests separated by spaces under the appropriate environment variable name
 setenv subtest_list_common     ""
-setenv subtest_list_non_replic "gtm8788 gtm7986 gtm8186 gtm8804 gtm8832"
+setenv subtest_list_non_replic "gtm8788 gtm7986 gtm8186 gtm8804 gtm8832 gtm8617"
 setenv subtest_list_replic     ""
 
 if ($?test_replic == 1) then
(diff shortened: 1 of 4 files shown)


v63003/instream.csh
```diff
@@ -19,13 +19,14 @@
 # gtm8186	    [vinay] Test DO, GOTO and ZGOTO can take offset without a label
 # gtm8804	    [vinay] Test zshow "t" produces only a summary of zshow "g" and "l"
 # gtm8832	    [vinay] Test string literal evaluating to >=1E47 produces a NUMOFLOW error
+# gtm8617	    [vinay] Tests functionality of STDNULLCOLL and NULL_SUBSCRIPTS
```
nars1: STDNULLCOLL -> MUPIP SET command for STDNULLCOLL

v63003/u_inref/gtm8617.csh
```diff
@@ -0,0 +1,51 @@
...
+echo "# Null collation set to GT.M"
+$gtm_tst/com/dbcreate.csh mumps 1 >>& db.txt
+$ydb_dist/mupip set -region DEFAULT -std
+$gtm_tst/com/dbcheck.csh >& db.txt
+
+echo "# Null collation set to M"
+$gtm_tst/com/dbcreate.csh mumps 1 >>& db.txt
+$ydb_dist/mupip set -region DEFAULT -nostd
+$gtm_tst/com/dbcheck.csh >& db.txt
+
```
nars1: Please remove the huge block of empty lines at the tail of this file.

v63003/u_inref/gtm8617.csh
```diff
@@ -0,0 +1,51 @@
...
+#	the license, please stop and do not read further.	#
+#								#
+#################################################################
+#
+# Case for each functionality described in the release note
+#
+
+echo "# Null subscript set to never"
+$gtm_tst/com/dbcreate.csh mumps 1 >>& db.txt
+$ydb_dist/mupip set -region DEFAULT -N=never
```
nars1: Please expand qualifier names in test scripts even though abbreviated forms are accepted. i.e. -N -> -NULL_SUBSCRIPTS, -std -> -stdnullcoll etc. in various places.

nars1: The output of the MUPIP SET command is included in the reference file. That is not enough proof that null subscripts is set to never. One needs to do a **dse dump -file |& grep "Null subscripts"** to verify that is indeed set in the db file header. Same with the other commands/qualifiers in this test.

v63003/u_inref/gtm8617.csh
```diff
@@ -0,0 +1,51 @@
...
+echo "# Null subscript set to existing"
+$gtm_tst/com/dbcreate.csh mumps 1 >>& db.txt
+$ydb_dist/mupip set -region DEFAULT -N=existing
+$gtm_tst/com/dbcheck.csh >& db.txt
+
+echo "# Null subscript set to always"
+$gtm_tst/com/dbcreate.csh mumps 1 >>& db.txt
+$ydb_dist/mupip set -region DEFAULT -N=always
+$gtm_tst/com/dbcheck.csh >& db.txt
+
```
nars1: Also add a test that NULLSUBS=<string> where string is none of the allowed 3 choices issues an error.

v63003/u_inref/gtm8617.csh
```diff
@@ -0,0 +1,51 @@
...
+echo "# Null collation set to GT.M"
+$gtm_tst/com/dbcreate.csh mumps 1 >>& db.txt
+$ydb_dist/mupip set -region DEFAULT -std
```
nars1: One needs to do a **dse dump -file |& grep "Standard Null Collation"** to verify that is indeed set in the db file header. Same with the other commands/qualifiers in this test.

PR #186, opened 2018-05-31
 v63003/gtm8735 (Tests GTM-8735 in V63003)

VinayYottaDB:

MUPIP SET -{FILE|REGION} recognizes the -[NO]READ_ONLY qualifier to indicate whether GT.M should treat an MM access method segment as read only for all users, including root. This designation augments UNIX authorizations and prevents any state updates that normally might require an operational action for a database with no current accessing (attached) processes. The GT.M help databases have -READ_ONLY set by default. Previously, a database such as the gtmhelp database in the GT.M distribution typically never received a data update but nevertheless could require a ROLLBACK, RECOVER or RUNDOWN to ensure a proper at-rest state. MUPIP emits an error on attempts to set -READ_ONLY on databases with the BG access method, or to set the access method to BG on databases with -READ_ONLY set. (GTM-8735)


v63003/u_inref/gtm8735.csh
```diff
@@ -0,0 +1,48 @@
...
+#
+#
+
+echo '# Creating database, manually setting variables so database is compatible with read only'
+setenv gtm_test_db_format "NO_CHANGE"
+setenv acc_meth MM
+setenv test_encryption NON_ENCRYPT
+source $gtm_tst/com/mm_nobefore.csh
+$gtm_tst/com/dbcreate.csh mumps 1 >>& create1.out
+$MUPIP SET -region DEFAULT -ACCESS_METHOD=MM -NOSTATS >>& settings.out
```
nars1: After every MUPIP SET command, I would like the following to be done to verify the db file header does say Read only status is what the MUPIP SET command said.

dse dump -file |& $grep "Read Only"

v63003/u_inref/gtm8735.csh
```diff
@@ -0,0 +1,48 @@
...
+$ydb_dist/mumps -run ^%XCMD "zwrite ^X"
+
+echo '# Setting access method to BG'
+$MUPIP SET -region DEFAULT -ACCESS_METHOD=BG
+
+echo '# Attempting to set default region to read only with a BG access method'
+$MUPIP SET -region DEFAULT -READ_ONLY
+
+$gtm_tst/com/dbcheck.csh mumps 1 >>& check1.out
+
```
nars1: I would like test of the following too

Verify that the help database files ($ydb_dist/*.dat) have the read_only flag set (use DSE DUMP -FILE to verify as well as attempting a SET and verify we get a DBPRIVERR error).

v63003/u_inref/gtm8735.csh
ksbhaskar: Since DSE opens and closes database files, might there be less overhead in running ``mupip dumpfhead -file <filename> | grep sgmnt_data.read_only`` or ``mupip dumpfhead -file <filename> | mumps -run %XCMD 'write:$zfind(%l,"sgmnt_data.read_only") %l,!``?

nars1: This test is a very simplistic one and we are going to do around 4 DSE invocations. It should be an insignificant change. Moreover, the user is more likely to run DSE to check the "Read Only" flag setting. And so it is better for us to use the same tool (DSE DUMP) instead of MUPIP DUMPFHEAD.

v63003/u_inref/gtm8735.csh
nars1: As discussed, I would like this done for all *.dat files in $ydb_dist, not just gtmhelp.dat. And also do a "ls -l " of $ydb_dist/*.dat and verify that owner/group/world all have read-only permissions (i.e. no one has write permissions).

v63003/outref/gtm8735.txt
```diff
@@ -24,8 +24,39 @@ Database file ##TEST_PATH##/mumps.dat now has BG access method
 %YDB-W-WCWRNNOTCHG, Not all specified database files were changed
   Access method                          BG  Global Buffers                1024
   LOCK shares DB critical section     FALSE  Read Only                      OFF
-# Displaying status of gtmhelp database
+# Displaying status of gtmhelp databases
```
nars1: Add at the end of the line : "to verify files have read-only permissions for user/group/other"

v63003/outref/gtm8735.txt
```diff
@@ -24,8 +24,39 @@ Database file ##TEST_PATH##/mumps.dat now has BG access method
...
+-r--r--r-- ##SOURCE_PATH##/dsehelp.dat
+-r--r--r-- ##SOURCE_PATH##/gdehelp.dat
+-r--r--r-- ##SOURCE_PATH##/gtmhelp.dat
+-r--r--r-- ##SOURCE_PATH##/lkehelp.dat
+-r--r--r-- ##SOURCE_PATH##/mupiphelp.dat
+#Displaying status of ##SOURCE_PATH##/dsehelp.gld
```
nars1: Space after # needed. #Displaying -> # Displaying

v63003/outref/gtm8735.txt
```diff
@@ -24,8 +24,39 @@ Database file ##TEST_PATH##/mumps.dat now has BG access method
...
+  LOCK shares DB critical section     FALSE  Read Only                       ON
+%YDB-E-DBFILOPERR, Error doing database I/O to region ##SOURCE_PATH##/dsehelp.dat
```
nars1: Need a comment line here describing what is being attempted (MUPIP SET). Same before the output where a SET is attempted.

v63003/u_inref/gtm8735.csh
```diff
@@ -60,15 +60,17 @@ $DSE dump -file|&$grep "Access method"
 $DSE dump -file|&$grep "Read Only"
 
 
-echo '# Displaying status of gtmhelp databases'
+echo '# Displaying status of gtmhelp databases to verify files have read-only permissions for user/group/other'
 ls -l $ydb_dist/*.dat|awk '{print $1,$9}'
```
nars1: Need to use $tst_awk (not awk) in test system scripts. Just like $grep (instead of grep)

v63003/u_inref/gtm8735.csh
```diff
@@ -60,15 +60,17 @@ $DSE dump -file|&$grep "Access method"
...
 
 set gbldir = `ls $ydb_dist/*.gld`
 foreach dir ( $gbldir:as/ / / )
-	echo "#Displaying status of "$dir
+	echo "# Displaying status of "$dir
 	setenv ydb_gbldir $dir
 	$DSE dump -file|&$grep "Read Only"
```
nars1: Spaces before and after |&

v63003/u_inref/gtm8735.csh
```diff
@@ -60,15 +60,17 @@ $DSE dump -file|&$grep "Access method"
 $DSE dump -file|&$grep "Read Only"
 
 
-echo '# Displaying status of gtmhelp databases'
+echo '# Displaying status of gtmhelp databases to verify files have read-only permissions for user/group/other'
 ls -l $ydb_dist/*.dat|awk '{print $1,$9}'
 
 set gbldir = `ls $ydb_dist/*.gld`
```
nars1: Lines 66 and 67 can be replaced as follows

foreach dir ($ydb_dist/*.gld)

PR #187, opened 2018-05-31
New v63004/gtm8923 subtest (tests GTM-8923 in V63004)

JakeYottaDB:

READ * and WRITE * operate correctly on files and sockets with a CHSET of UTF-16, UTF-16BE, or UTF-16LE. Previously READ * returned an incorrect codepoint and WRITE * produced an incorrect character or a BADCHAR error. (GTM-8923)


v63004/instream.csh
```diff
@@ -26,13 +26,14 @@
...
 # gtm8643	    [jake]  Tests that YDB no longer enforces queue depth limit of 5 for SOCKET devices.
+# gtm8923	    [jake]  Tests the READ * and WRITE * commands no longer produce errors or incorrect output for files or sockets with CHSET={UTF-16,UTF-16BE,UTF-16LE}
 
 echo "v63004 test starts..."
 
 # List the subtests separated by spaces under the appropriate environment variable name
 setenv subtest_list_common     ""
 setenv subtest_list_non_replic ""
-setenv subtest_list_non_replic "$subtest_list_non_replic gtm8914 gtm8909 gtm8874 gtm8860 gtm8791 gtm8699 gtm8202 gtm1042 gtm8891 gtm8894 gtm8643"
+setenv subtest_list_non_replic "$subtest_list_non_replic gtm8914 gtm8909 gtm8874 gtm8860 gtm8791 gtm8699 gtm8202 gtm1042 gtm8891 gtm8894 gtm8643 gtm8923"
```
nars1: Can you split this into 2 lines. It is too long. We usually have a limit of 132 columns.

v63004/inref/gtm8923.m
```diff
@@ -0,0 +1,115 @@
...
+;	This source code contains the intellectual property	;
+;	of its copyright holder(s), and is made available	;
+;	under a license.  If you do not know the terms of	;
+;	the license, please stop and do not read further.	;
+;								;
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;
+gtm8923
+
+	SET encodings="UTF-16"
```
nars1: This seems like a redundant line.

v63004/inref/gtm8923.m
```diff
@@ -0,0 +1,115 @@
...
+	. USE file:CHSET=chset
+	. WRITE "1",!
+	. CLOSE file
+	. OPEN file:CHSET=chset:10
+	. USE file:CHSET=chset
+	. READ *x,!
+	. CLOSE file
+	. USE $P
+	. WRITE "READ * test on "_chset_" file:"
+	. WRITE x,!
```
nars1: Can you also do a ZWRITE x. That way we will know what undisplayable bytes if any it contains. Same comment applies to multiple places in this file.

v63004/inref/gtm8923.m
```diff
@@ -0,0 +1,115 @@
...
+
+	SET encodings="UTF-16"
+	SET encodings="UTF-16 UTF-16BE UTF-16LE"
+
+	FOR I=1:1:$L(encodings," ") DO
+	. SET chset=$P(encodings," ",I)
+	. SET file="gtm8923"_chset_".file"
+	.
+	. ;READ * test
+	. OPEN file:CHSET=chset:10
```
nars1: While having a timeout of 10 seconds in the open command is good, we don't handle the case where the open fails with a timeout. So it is better to remove the :10. If the open command hangs, we will know. But that way we don't need to handle timeout. Same comment applies to multiple places.

v63004/inref/gtm8923.m
```diff
@@ -0,0 +1,115 @@
...
+	. WRITE "1",!
+	. CLOSE file
+	. OPEN file:CHSET=chset:10
+	. USE file:CHSET=chset
+	. READ *x,!
+	. CLOSE file
+	. USE $P
+	. WRITE "READ * test on "_chset_" file:"
+	. WRITE x,!
+	. ;CLOSE file
```
nars1: Line can be removed.

v63004/inref/gtm8923.m
```diff
@@ -0,0 +1,115 @@
...
+	. USE file:CHSET=chset
+	. READ x,!
+	. CLOSE file
+	. USE $P
+	. WRITE "WRITE * test on "_chset_" file:"
+	. WRITE x,!
+
+        do ^job("child^gtm8923",2,"""""")     ; start 2 jobs ; .
+
+	WRITE ^readTests("1"),!
```
nars1: Can you use a loop for lines 51-58

nars1:
@JakeYottaDB : I verified that $char(174) (the Registered trademark symbol) does produce a BADCHAR error with V63003. So you can use that in addition to the $char(49) you already are testing with. That should cover all the cases mentioned in the release note.

v63004/outref/gtm8923.txt
```diff
@@ -1,19 +1,61 @@
 setenv gtm_chset UTF-8
 unset echo
-# Testing VIEW "[NO]STATSHARE"[:<region-list>]
+# Testing READ * and WRITE * commands on sockets and files of various UTF CHSETs
 # Create a single region DB with region DEFAULT
 
 # Run gtm8923.m to test WRITE * and READ * commands
-READ * test on UTF-16 file:49
```
nars1: I like the previous output of each test case in one line instead of the current one with 4 or 5 lines per test case. Can you please fix the test so that output format is preserved.

PR #188, opened 2018-06-01
New v63003/gtm8767 subtest (Tests GTM-8767 in V63003)

VinayYottaDB:

MUPIP SET for file or region accepts -H[ARD_SPIN_COUNT]=<integer count> and -SPIN[_SLEEP_MASK]=<hexadecimal mask>; previously it did not support changes to the hard spin count and required -SPIN_SLEEP_LIMIT to change the spin sleep mask. MUPIP SET no longer supports the -SPIN_SLEEP_LIMIT qualifier.(GTM-8767)


v63003/u_inref/gtm8767.csh
```diff
@@ -0,0 +1,35 @@
...
+#	This source code contains the intellectual property	#
+#	of its copyright holder(s), and is made available	#
+#	under a license.  If you do not know the terms of	#
+#	the license, please stop and do not read further.	#
+#								#
+#################################################################
+#
+#
+set randint = $$
+set randhex = `openssl rand -hex 3`
```
nars1: Please use test system script genrandnumbers.csh to generate random numbers.

v63003/u_inref/gtm8767.csh
```diff
@@ -0,0 +1,35 @@
...
+set randnum = $$
+
+$gtm_tst/com/dbcreate.csh mumps 1 >>& create1.out
+echo "Setting Hard Spin Count"
+$MUPIP SET -region DEFAULT -HARD_SPIN_COUNT=$randint
+#Will only display if Hard Spin Count is correct in DSE Dump
```
nars1: Remove this line and instead add a line like

echo "# Verify the random hard spin count shows up in DSE DUMP -FILE"

Also do similar change in other places where DSE DUMP -FILE is invoked.

v63003/u_inref/gtm8767.csh
```diff
@@ -0,0 +1,35 @@
...
+#								#
+#################################################################
+#
+#
+set randint = $$
+set randhex = `openssl rand -hex 3`
+set randnum = $$
+
+$gtm_tst/com/dbcreate.csh mumps 1 >>& create1.out
+echo "Setting Hard Spin Count"
```
nars1: AReplace this line with

echo "# Setting Hard Spin Count to a random value using MUPIP SET -HARD_SPIN_COUNT"

Also do similar change in other places where MUPIP SET is invoked.

v63003/u_inref/gtm8767.csh
```diff
@@ -0,0 +1,35 @@
...
+#								#
+#	This source code contains the intellectual property	#
+#	of its copyright holder(s), and is made available	#
+#	under a license.  If you do not know the terms of	#
+#	the license, please stop and do not read further.	#
+#								#
+#################################################################
+#
+#
+set randint = $$
```
nars1: As discussed, this is not a random number and should be replaced with a genrandnumbers.csh call.
Additionally, I would like a hard_spin_count of 0 and 2147483647 also tested as being allowed and a value of 2147483648 tested as being not allowed.

PR #190, opened 2018-06-01
New v63004/gtm8903 subtest (tests GTM-8903 in V63004)

JakeYottaDB:

$SELECT() deals with cases where the first value is evaluated to a literal TRUE (1) and later arguments to the function 
contain one or more global references. Due to a regression in V6.3-002 associated with compiler optimizations 
caused this combination to produce a SIG-11 (segmentation violation) at compilation. (GTM-8903)

NOTE:
    With the information provided in the patch notes, a sig-11 could  not be reproduced. The test instead is designed to produce a GTMASSERT2 in regressed releases.


v63004/inref/gtm8903.m
```diff
@@ -0,0 +1,22 @@
...
+;								;
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;
+gtm8903
+	;We'll set some local and global variables to reference later
+	SET x=1
+	SET ^x=2
+
+	;$S[ELECT](tvexpr:expr[,...])
+	WRITE "0!$SELECT(1:x&x,1:^x),!",!
```
nars1: Duplication of the $SELECT expression can be avoided by setting a variable and using the xecute command.

set xstr="WRITE ...." write xstr," = "  xecute xstr

v63004/instream.csh
```diff
@@ -27,13 +27,14 @@
 # gtm8894	    [vinay] Tests that $zreldate outputs in the form YYYYMMDD 24:60
 # gtm8643	    [jake]  Tests that YDB no longer enforces queue depth limit of 5 for SOCKET devices.
 # gtm8922	    [jake]  Tests the functionality of VIEW keywords that take <region-list> expressions
+# gtm8903	    [jake]  Tests $S[ELECT](<TRUE EXPR>:<expr>,<EXPR>:<GLOBAL REFERENCE>) function for errors
```
nars1: No need to have $S[ELECT] in test. That is needed in documentation to indicate various choices for the same function. $SELECT is fine here.

The test is actually for $SELECT(1:<boolean-expression>,<EXPR>:<GLOBAL REFERENCE>)

v63004/u_inref/gtm8903.csh
```diff
@@ -0,0 +1,24 @@
...
+#								#
+#################################################################
+#
+echo '# Testing $S[ELECT](<TRUE>:<expr>,<GLBL references>)'
+echo ''
+
+echo "# Create a DB with region DEFAULT"
+$gtm_tst/com/dbcreate.csh mumps 1 >>& dbcreate_log_1.txt
+
+echo '# Run gtm8903.m to'
```
nars1: Sentence seems unfinished

PR #192, opened 2018-06-01
 New v63003/gtm8779 (Tests GTM-8779 in V63003)

VinayYottaDB:

MUPIP FREEZE sends a DBFREEZEON/DBFREEZEOFF message to the system log for each region whose freeze state is changed. (GTM-8779)


v63003/u_inref/gtm8779.csh
```diff
@@ -0,0 +1,44 @@
...
+
+echo "# Setting Default Region to UNFROZEN"
+$MUPIP FREEZE -OFF DEFAULT
+set t3 = `date +"%b %e %H:%M:%S"`
+echo "# Verifying System received a DBFREEZEOFF message"
+$gtm_tst/com/getoper.csh "$t2" "$t3" t2t3.txt
+
+cat t2t3.txt |& $grep "DBFREEZEOFF" | $tst_awk '{print $6}'
+
+$gtm_tst/com/dbcheck.csh >>& check.out
```
nars1: Instead of the above, I would like to have 3 regions AREG, BREG and DEFAULT. And run the below two test scenarios.

1) Have AREG freeze set to off, BREG freeze set to ON and DEFAULT freeze set to off. And do a MUPIP FREEZE -ON "*" to verify that DBFREEZEON messages get sent only for AREG and DEFAULT and that no DBFREEZEOFF messages get sent.

2) Have AREG freeze set to off, BREG freeze set to ON and DEFAULT freeze set to on. And do a MUPIP FREEZE -OFF "*" and verify DBFREEZEOFF only gets sent for BREG and DEFAULT and no DBFREEZEON messages get sent.

v63003/u_inref/gtm8779.csh
```diff
@@ -0,0 +1,44 @@
...
+# Sleep commands to ensure a difference in time of at least 1 second so getoper.csh
+# will catch the message every time
+sleep 1
+$MUPIP FREEZE -ON DEFAULT
+set t2 = `date +"%b %e %H:%M:%S"`
+sleep 1
+echo "# Verifying System received a DBFREEZEON message"
+$gtm_tst/com/getoper.csh "$t1" "$t2" t1t2.txt
+
+cat t1t2.txt |& $grep "DBFREEZEON" | $tst_awk '{print $6}'
```
nars1: We need to capture the full DBFREEZEON and any DBFREEZEOFF messages here (search for DBFREEZE*).

Also not just the last column ($6). That way we know which region the message corresponds to.

v63003/outref/gtm8779.txt
```diff
@@ -1,18 +1,38 @@
...
-# Resetting to initial state
+
+# Resetting DEFAULT and AREG to OFF, BREG to ON
+Region DEFAULT is now UNFROZEN
+All requested regions unfrozen
+Region AREG is now UNFROZEN
+All requested regions unfrozen
+Region BREG is now FROZEN
+All requested regions frozen
+# Verifying System received a DBFREEZEOFF message for only AREG and BREG
```
nars1: BREG -> DEFAULT

v63003/u_inref/gtm8779.csh
```diff
@@ -15,49 +15,80 @@
 #
 
 $gtm_tst/com/dbcreate.csh mumps 3 >>& create.out
+echo "# Initially, all regions set to OFF"
```
nars1: The echo should happen AFTER the "if ($status)" check. Since $status from the dbcreate call gets reset after the echo call.

PR #201, opened 2018-06-06
Changes to v63003/gtm8735

VinayYottaDB:
Added loop to account for the occasional different error

### DIFF
--- a/v63003/outref/gtm8735.txt
+++ b/v63003/outref/gtm8735.txt
@@ -33,40 +33,40 @@ Database file ##TEST_PATH##/mumps.dat now has BG access method
 # Displaying status of ##SOURCE_PATH##/dsehelp.gld
   LOCK shares DB critical section     FALSE  Read Only                       ON
 # Attempting to change to no read only
-%YDB-E-DBFILOPERR, Error doing database I/O to database file ##SOURCE_PATH##/dsehelp.dat
+%YDB-E-DBFILOPERR, Error doing database I/O to database file ##SOURCE_REALPATH##/dsehelp.dat
 %SYSTEM-E-ENO13, Permission denied
 %YDB-E-WCERRNOTCHG, Not all specified database files were changed
 # Attempting to write to database
-%YDB-E-DBPRIVERR, No privilege for attempted update operation for file: ##SOURCE_PATH##/dsehelp.dat
+%YDB-E-DBPRIVERR, No privilege for attempted update operation for file: ##SOURCE_REALPATH##/dsehelp.dat
 # Displaying status of ##SOURCE_PATH##/gdehelp.gld
   LOCK shares DB critical section     FALSE  Read Only                       ON
 # Attempting to change to no read only
-%YDB-E-DBFILOPERR, Error doing database I/O to database file ##SOURCE_PATH##/gdehelp.dat
+%YDB-E-DBFILOPERR, Error doing database I/O to database file ##SOURCE_REALPATH##/gdehelp.dat
 %SYSTEM-E-ENO13, Permission denied
 %YDB-E-WCERRNOTCHG, Not all specified database files were changed
 # Attempting to write to database
-%YDB-E-DBPRIVERR, No privilege for attempted update operation for file: ##SOURCE_PATH##/gdehelp.dat
+%YDB-E-DBPRIVERR, No privilege for attempted update operation for file: ##SOURCE_REALPATH##/gdehelp.dat
 # Displaying status of ##SOURCE_PATH##/gtmhelp.gld
   LOCK shares DB critical section     FALSE  Read Only                       ON
 # Attempting to change to no read only
-%YDB-E-DBFILOPERR, Error doing database I/O to database file ##SOURCE_PATH##/gtmhelp.dat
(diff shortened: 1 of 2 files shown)


v63003/outref/gtm8735.txt
```diff
@@ -33,40 +33,40 @@ Database file ##TEST_PATH##/mumps.dat now has BG access method
 # Displaying status of ##SOURCE_PATH##/dsehelp.gld
   LOCK shares DB critical section     FALSE  Read Only                       ON
 # Attempting to change to no read only
-%YDB-E-DBFILOPERR, Error doing database I/O to database file ##SOURCE_REALPATH##/dsehelp.dat
+%YDB-E-DBFILOPERR, Error doing database I/O to region ##SOURCE_PATH##/dsehelp.dat
```
nars1: This change seems to be a regression. Why was this necessary?

v63003/u_inref/gtm8735.csh
```diff
@@ -66,16 +66,23 @@ $DSE dump -file|&$grep "Read Only"
...
+set i = 0
 foreach dir ($ydb_dist/*.gld)
 	echo "# Displaying status of "$dir
 	setenv ydb_gbldir $dir
 	$DSE dump -file |& $grep "Read Only"
 	echo "# Attempting to change to no read only"
-	$MUPIP SET -region DEFAULT -NOREAD_ONLY
+	set x = ""
+	while ("$x" == "")
+		@ i++
```
nars1: We don't need an "i" variable for this. Use temp.out for all output instead of temp$i.out. Also if a READONLYLKFAIL error is seen, then redo the loop. Else break out of the loop. This way if we have any error other than READONLYLKFAIL or DBFILOPERR we still break out of the loop instead of incorrectly continuing the loop.

v63003/u_inref/gtm8735.csh
```diff
@@ -66,16 +66,23 @@ $DSE dump -file|&$grep "Read Only"
...
 $gtm_tst/com/lsminusl.csh $ydb_dist/*.dat|$tst_awk '{print $1,$9}'
-
+set i = 0
 foreach dir ($ydb_dist/*.gld)
 	echo "# Displaying status of "$dir
 	setenv ydb_gbldir $dir
 	$DSE dump -file |& $grep "Read Only"
 	echo "# Attempting to change to no read only"
-	$MUPIP SET -region DEFAULT -NOREAD_ONLY
+	set x = ""
```
nars1: There needs to be a comment on why a loop is needed i.e. what error (READONLYLKFAIL) is possible.

v63003/outref/gtm8735.txt
```diff
@@ -33,40 +33,40 @@ Database file ##TEST_PATH##/mumps.dat now has BG access method
 # Displaying status of ##SOURCE_PATH##/dsehelp.gld
   LOCK shares DB critical section     FALSE  Read Only                       ON
 # Attempting to change to no read only
-%YDB-E-DBFILOPERR, Error doing database I/O to database file ##SOURCE_REALPATH##/dsehelp.dat
+%YDB-E-DBFILOPERR, Error doing database I/O to database file ##SOURCE_PATH##/dsehelp.dat
```
nars1: This was SOURCE_REALPATH before but is now SOURCE_PATH. It has to be the former or else this test will fail on systems where the test output directory is a soft link.

v63003/u_inref/gtm8735.csh
```diff
@@ -66,13 +66,22 @@ $DSE dump -file|&$grep "Read Only"
...
 foreach dir ($ydb_dist/*.gld)
 	echo "# Displaying status of "$dir
 	setenv ydb_gbldir $dir
 	$DSE dump -file |& $grep "Read Only"
 	echo "# Attempting to change to no read only"
-	$MUPIP SET -region DEFAULT -NOREAD_ONLY
+	$MUPIP SET -region DEFAULT -NOREAD_ONLY >>& temp.out
+	# Occasionally there is a READONLYLKFAIL Error instead of the desired DBFILEOPERR. We have the program retry
+	# if it encounters this
+	set x = `cat temp.out |& $grep READONLYLKFAIL`
```
nars1: Please avoid duplication in line 77 and 81. You can use a "while (1)" (i.e. infinite loop) construct.

Also, the grep can be simplified by doing "$grep READONLYLKFAIL temp.out > /dev/null" and checking "if ($status)" in the next line. $status will be 0 if a match occurs and non-zero otherwise.

v63003/u_inref/gtm8735.csh
```diff
@@ -71,14 +71,16 @@ foreach dir ($ydb_dist/*.gld)
...
 	$DSE dump -file |& $grep "Read Only"
 	echo "# Attempting to change to no read only"
-	$MUPIP SET -region DEFAULT -NOREAD_ONLY >>& temp.out
 	# Occasionally there is a READONLYLKFAIL Error instead of the desired DBFILEOPERR. We have the program retry
 	# if it encounters this
-	set x = `cat temp.out |& $grep READONLYLKFAIL`
-	while ("$x" != "")
-		rm temp.out
+	while (1)
 		$MUPIP SET -region DEFAULT -NOREAD_ONLY >>& temp.out
```
nars1: >>& --> >&
And avoid use of "x" variable by using $status.

PR #206, opened 2018-06-06
New v63004/gtm3146 subtest (tests GTM-3146 in V63004) 

JakeYottaDB:
                                                                                                                                                                                          
GTM ensures that alias and path settings do not interfere with the intentions of the actions 
it takes using invocations of the POSIX system() function. This issue was only observed in 
the GTM development environment, and was never reported by a user. (GTM-3146)


v63004/outref/gtm3146.txt
```diff
@@ -0,0 +1,20 @@
+# Testing 
+# Create a 1 region DB region DEFAULT
+
+# Create a file, shell.csh, to be sourced by child shells : 
+
+cat shell.csh
+#!/usr/local/bin/tcsh -f
+set echo ; set verbose
+alias cp "jakewashere"
```
nars1: Add a comment here that the MUPIP BACKUP done later (which internally does a "cp") used to fail in prior versions due to the "cp" alias here.

v63004/u_inref/gtm3146.csh
```diff
@@ -0,0 +1,59 @@
...
+echo 'cat shell.csh'
+cat ./shell.csh
+echo ''
+
+echo '# Set $SHELL env var to have child shell shell.csh '
+setenv SHELL "$PWD/shell.csh"
+echo ''
+
+echo "# Running Mupip with "'$SHELL'" ./shell.csh"
+$MUPIP BACKUP "*" tstBackup >>& backup.log
```
nars1: Check $status after MUPIP BACKUP to ensure it is 0. If not, print an error.

v63004/outref/gtm3146.txt
JakeYottaDB: The comment has been added to the generated shell.csh file. This way the comment will appear right below the cp alias when shell.csh is cat-ed to the screen.

v63004/u_inref/gtm3146.csh
JakeYottaDB: A $status check block, mirroring those used after dbcreate and dbcheck, has been added after the $MUPIP BACKUP call.

PR #214, opened 2018-06-07
New v63003/gtm8787 (tests GTM-8787 in V63003)

VinayYottaDB:

MUPIP JOURNAL -EXTRACT='-stdout' appropriately handles its termination; previously, if stdout was already closed this specification produced a segmentation violation (SIG-11). (GTM-8787)


v63003/u_inref/gtm8787.csh
```diff
@@ -0,0 +1,57 @@
...
+# allowing the system to easily exit while extracting
+$ydb_dist/mumps -run ^%XCMD "for i=1:1:10 set ^x(i)=i"
+set t = `date +"%b %e %H:%M:%S"`
+
+echo "# Running extract and terminating in separate terminal"
+(expect -d $gtm_tst/$tst/u_inref/gtm8787.exp > expect.outx) >& xpect.dbg
+if ($status) then
+	echo "EXPECT FAILED"
+endif
+$ydb_dist/mumps -run gtm8787 >>& temp.out
```
nars1: Now that we are using this logic a lot more, I would like a framework change that is much better. And that is to enhance getoper.csh so if it is passed "" as $2 (the after_time), it does this random string generation and waits for that to show up). That way this test (and your prior tests too) can use this framework rather than doing "mumps -run gtm8787" etc.

v63003/u_inref/gtm8787.exp
```diff
@@ -0,0 +1,34 @@
...
+}
+
+
+# Have columns higher than 80 as that can cause test failures on lines that are just above 80 columns in length
+send -- "stty cols 132\r"
+expect "stty cols 132\r"
+
+expect -exact ">"
+send -- "\$MUPIP Journal -Extract=-stdout -forward mumps.mjl & \r"
+expect -exact "x(10)"
```
nars1: We are doing just 10 updates and expecting to see x(10) before exiting the expect script (i.e. Xing the terminal). I would prefer we do 1000 updates instead of 10. But please confirm that with this, we are still able to reproduce the KILLBYSIG error in V63002.

v63003/u_inref/gtm8787.csh
```diff
@@ -14,40 +14,35 @@
...
-source $gtm_tst/com/gtm_test_setbgaccess.csh
-source $gtm_tst/com/gtm_test_setbeforeimage.csh
+echo "# Creating database,journal file"
 $gtm_tst/com/dbcreate.csh mumps 1>>& create.out
 if ($status) then
 	echo "DB Create Failed, Output Below"
 	cat create.out
 endif
-$MUPIP Set -Region Default -Journal=enable,on,before,file=mumps.mjl >>& jnlcreate.out
+$MUPIP Set -Region Default -Journal=enable,on,nobefore,file=mumps.mjl >>& jnlcreate.out
```
nars1: Add a comment that we use nobefore so this test can work with BG or MM access method (randomly chosen by test framework).

v63003/u_inref/gtm8787.csh
```diff
@@ -14,40 +14,35 @@
...
 
 echo "# Running extract and terminating in separate terminal"
 (expect -d $gtm_tst/$tst/u_inref/gtm8787.exp > expect.outx) >& xpect.dbg
 if ($status) then
 	echo "EXPECT FAILED"
 endif
-$ydb_dist/mumps -run gtm8787 >>& temp.out
-set s = `cat temp.out`
-$gtm_tst/com/getoper.csh "$t" "" log.outx "" $s
+$gtm_tst/com/getoper.csh "$t"
```
nars1: I would prefer a comment at the end of the line indicating the syslog messages will be logged to syslog.txt.

v63003/u_inref/gtm8787.csh
```diff
@@ -14,40 +14,35 @@
...
 
 echo "# Searching Sys Log for a KILLBYSIG Error (Expecting nothing, would be found in previous versions)"
-cat log.outx |& $grep KILLBYSIG |& $tst_awk '{print $5 " " $6 " " $7 " " $8 " " $9 " " $10 " " $11 " " $12}'
+cat syslog.txt |& $grep KILLBYSIG |& $tst_awk '{print $6 " " $7 " " $8 " " $9 " " $10 " " $11 " " $12}'
```
nars1: The space and " " are not needed here and in a later similar usage. Can be replaced by a comma (i.e. $6,$7,...)

v63003/u_inref/gtm8787.csh
nars1: Now that you have enhanced getoper.csh to take care of this, please undo gtm8779.m too (just like you undid gtm8787.m). Thanks.

PR #216, opened 2018-06-08
Enhance test framework to capture syslog/ps/ipc information at time of both TEST-E-HANG and TEST-E-TIMEDOUT alerts

nars1:


com/gtm_test_watchdog.csh
```diff
@@ -82,18 +82,41 @@ while !( -e $exit_file)
 		set runtime =  `echo $timestart.$timenow | $tst_awk -F \. -f $gtm_tst/com/diff_time.awk`
 		if ( ( $runtime >= $gtm_test_hang_alert_sec ) && (! $?mailsent) ) then
 			# The test has been running for longer than $gtm_test_hang_alert_sec.
+			# Send a TEST-E-HANG email
+			#
+			cd $cursubtestdir
+			# Capture ps/ipcs etc. at time of TEST-E-HANG report
+			$gtm_tst/com/capture_ps_ipcs_netstat_lsof.csh	>& capture_ps_ipcs_netstat_lsof_HANG.out
```
estess: Is it possible for this to run more than once and if so is it appropriate for it to overwrite the output file?

nars1: It is guaranteed to not run more than once by the check in line 83 **&& (! $?mailsent)** as that gets set to 1 at the end when we run this block of code the first time around.

PR #222, opened 2018-06-11
New v63003/gtm8856 subtest (tests GTM-8856 in V63003)

VinayYottaDB:

GT.M defers literal optimizations involving patterns within an XECUTE as well as evaluations that encounter issues with the pattern table. Due to a regression associated with GTM-8573 in V6.3-001[A], these combinations caused an abnormal termination with a segmentation violation (SIG-11). (GTM-8856)


v63003/instream.csh
```diff
@@ -31,14 +31,15 @@
 # gtm8780	    [vinay] Tests $SELECT() produces a syntax error for an omitted colon after a literal true argument
 # gtm8787	    [vinay] Tests Mupip Journal -Extract=-stdout properly handles its termination
 # gtm8889	    [vinay] Tests zhelp does not produce an error when <Ctrl C> is pressed
+# gtm8856	    [vinay] Tests XECUTE+evaluations that encounter issues with a pattern table produce an error while compiling
```
nars1: Replace above line with

Tests literal optimizations that can result in errors at compile time are deferred if done inside XECUTE

PR #234, opened 2018-06-13
New v63003/gtm8794 subtest (Tests GTM-8794 in V63003)

VinayYottaDB:

Copies of a database file made while a MUPIP FREEZE -ONLINE -ON is in effect can be used on the same system by performing a MUPIP RUNDOWN -OVERRIDE and a MUPIP FREEZE -OFF on the copy. Previously, an attempt to remove the freeze on the copy would attempt to modify the journal files of the original database and fail. Note that this change moved the ^%PEEKBYNAME item "sgmnt_data.freeze_online" to "node_local.freeze_online"; for this release (V6.3-003) only, ^%PEEKBYNAME recognizes either designation, but going forward, it will not. If you have code referencing this item, please revise it. (GTM-8794)


v63003/instream.csh
```diff
@@ -46,7 +47,7 @@ echo "v63003 test starts..."
 setenv subtest_list_common     ""
 setenv subtest_list_non_replic "gtm8788 gtm7986 gtm8186 gtm8804 gtm8832 gtm8617 gtm4212 gtm8732nr gtm8767 gtm8735 gtm8779 gtm8798"
 setenv subtest_list_non_replic "$subtest_list_non_replic gtm8846 gtm8780 gtm8787 gtm8889 gtm8856 gtm8857 gtm8854 gtm8795 gtm8839"
-setenv subtest_list_non_replic "$subtest_list_non_replic gtm8781 gtm8849"
+setenv subtest_list_non_replic "$subtest_list_non_replic gtm8781 gtm8849 gtm8794"
 setenv subtest_list_replic     "gtm8732r gtm8795"
```
nars1: Need to include this test in the replic list too.

v63003/u_inref/gtm8794.csh
```diff
@@ -0,0 +1,53 @@
...
+echo ""
+echo "# Performing Rundown on mumps.dat in temp folder"
+$MUPIP RUNDOWN -OVERRIDE -FILE mumps.dat
+
+echo ""
+echo "# Turning off Freeze"
+$MUPIP FREEZE -OFF "*"
+
+echo "# Updating Database"
+# Journaling must be off in order for us to update database
```
nars1: Journaling does not need to be off. It just needs to be point to a journal file that is different from the source journal file. So please randomly switch journal files using MUPIP SET JOURNAL=ON here that way we verify that and NOJOURNAL scenarios work.

v63003/u_inref/gtm8794.csh
```diff
@@ -0,0 +1,53 @@
...
+cp mumps.dat temp/mumps.dat
+cp mumps.gld temp/mumps.gld
+cp mumps.mjl temp/mumps.mjl
+cd temp
+echo ""
+echo "# Performing Rundown on mumps.dat in temp folder"
+$MUPIP RUNDOWN -OVERRIDE -FILE mumps.dat
+
+echo ""
+echo "# Turning off Freeze"
```
nars1: Freeze -> Freeze on mumps.dat in temp folder

VinayYottaDB: The freeze is applied to the region, not the database folder

nars1: That is fine. But the region in turn points to a database file and it is helpful to know the freeze off is being done on the copy database file and not the live one.

PR #237, opened 2018-06-15
Fixed subtest failures v63003/gtm8794 for replic tests

VinayYottaDB:


v63003/u_inref/gtm8794.csh
```diff
@@ -37,8 +38,14 @@ $MUPIP FREEZE -OFF DEFAULT
 
 echo "# Updating Database"
 # Reassigning jnl file for the database to be in the temp folder (Randomly set to before/nobefore by the tests framework)
-$MUPIP SET -REGION DEFAULT -JOURNAL=disable -REPLICATION=off>&jnl.outx
-$MUPIP SET -REGION DEFAULT $tst_jnl_str -REPLICATION=on>&jnl.outx
+$MUPIP SET -REGION DEFAULT -JOURNAL=disable -REPLICATION=OFF>&jnl.outx
+if($status) then
+	echo "Bad MUPIP SET"
```
nars1: MUPIP SET -> MUPIP SET JOURNAL

Line 43 and 47 currently have same error message so one will have no clue which line had the error.

v63003/u_inref/gtm8794.csh
```diff
@@ -37,8 +38,14 @@ $MUPIP FREEZE -OFF DEFAULT
 
 echo "# Updating Database"
 # Reassigning jnl file for the database to be in the temp folder (Randomly set to before/nobefore by the tests framework)
-$MUPIP SET -REGION DEFAULT -JOURNAL=disable -REPLICATION=off>&jnl.outx
-$MUPIP SET -REGION DEFAULT $tst_jnl_str -REPLICATION=on>&jnl.outx
+$MUPIP SET -REGION DEFAULT -JOURNAL=disable -REPLICATION=OFF>&jnl.outx
```
nars1: 1) jnl.outx -> setjnl.out

outx means all errors here would be ignored by the test system framework. We don't want that unless we are expecting an error (which this case is not in my understanding).

2) Also, spaces before and after >&

v63003/u_inref/gtm8794.csh
```diff
@@ -37,8 +38,14 @@ $MUPIP FREEZE -OFF DEFAULT
...
+if($status) then
+	echo "Bad MUPIP SET"
+endif
+$MUPIP SET -REGION DEFAULT $tst_jnl_str >&jnl.outx
```
nars1: jnl.outx -> mupipset.out

We do not want two commands going to the same file without append (lines 41 and 45). One will overwrite the other's output.

PR #239, opened 2018-06-15
 New v63003/gtm8587 subtest (Tests GTM-8587 in V63003)

VinayYottaDB:

GT.M sets $KEY to the characters terminating a READ, and NULL if terminated otherwise (e.g. FIX format, end of file, or timeout). When it encounters an error during an I/O, GT.M sets $DEVICE to "1," followed by an error description. Previously, GT.M did not maintain $KEY for sequential devices and only maintained $DEVICE for certain I/O errors. (GTM-8587)


v63003/outref/gtm8587.txt
```diff
@@ -0,0 +1,10 @@
+# RUNNING TEST FOR $KEY
+# Check $DEVICE to see it is 0 and $KEY contains the terminating character
+$DEVICE=0
+$KEY=     (In previous versions, this would be an empty string)
```
nars1: It looks like an empty string even in the current version. If it is a control character, need a zwrite instead of write so we can see exactly what it has.

v63003/outref/gtm8587.txt
```diff
@@ -0,0 +1,10 @@
...
+
+# RUNNING TEST FOR $DEVICE
```
nars1: $DEVICE -> $DEVICE maintenance for files in case of errors

v63003/outref/gtm8587.txt
```diff
@@ -0,0 +1,10 @@
+# RUNNING TEST FOR $KEY
```
nars1: I would like test cases that show $key as null and as non-null (like the release note states).

v63003/u_inref/gtm8587.csh
```diff
@@ -16,7 +16,12 @@
...
+echo "# Triggering an IO Permissions Error, expecting a message in $device (would be zero in previous versions)"
+$ydb_dist/mumps -run dollardevice^gtm8587
+echo ""
 echo "# RUNNING TEST FOR $key\rIF YOU SEE THIS THE TEST FAILED">>&temp1.txt
+echo "# Ending a read with a new line control character, expecting $key to reflect this (would be null in previous versions)"
 $ydb_dist/mumps -run dollarkey^gtm8587 temp1.txt
-echo "# RUNNING TEST FOR $device"
-$ydb_dist/mumps -run dollardevice^gtm8587
+echo ""
+echo "# Ending a read with an EOF, expecting $key to be null"
```
nars1: The echo says we are ending a read with an EOF but we are not passing any parameter to dollarkey^gtm8587. So what file does it open? Is it really testing read EOF.

VinayYottaDB: It is not opening a file, when a READ encounters an empty file it terminates as an EOF (I believe the website mentions this). I provided $device to confirm that the read was terminated by this, but I can change it so it reads an actual file and gets terminated by an eof if you would prefer that

nars1: Agree $DEVICE is confirmation enough. Thanks. No more changes needed.

PR #245, opened 2018-06-19
Change to v62000/gtm8086 subtest (tests GTM-8699 in V63004) 

JakeYottaDB:
Added the following check:

Previously the [PREVJNLLINKCUT error] message could be reported in rare cases where a journal switch was deferred, then later switched without cutting links

### DIFF
--- a/v62000/u_inref/gtm8086.csh
+++ b/v62000/u_inref/gtm8086.csh
@@ -59,6 +59,7 @@ if ($?test_replic) then
 	# has jfh->is_not_latest_jnl set to TRUE.
 	set jnlswitchretry = `$gtm_exe/mumps -run rand 2`
 	if ($jnlswitchretry) then
+		set jnlswitchretry_time = `date +"%b %e %H: %M: %S"`
 		$MSR STOPRCV INST1 INST2 >& jnlswitchretry_1.log
 	endif
 else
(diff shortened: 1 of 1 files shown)


v62000/u_inref/gtm8086.csh
```diff
@@ -145,4 +146,13 @@ while ($jobnum < $gtm_test_jobcnt)
 	$grep -vE 'JNLEXTEND|JNLCLOSE|NOTALLDBRNDWN|GVRUNDOWN' impjob_imptp1.xmje${jobnum}x > impjob_imptp1.mje${jobnum}
 end
 
+if ($?test_replic) then
+	if ($jnlswitchretry) then
+		# getoper will capture syslog messages generated since starting the jnlswitchretry error
+		$gtm_tst/com/getoper.csh "$jnlswitchretry_time"
+		# Confirm there is no PREVJNLLINKCUT error in the output
```
nars1: Add a comment here too about "tests GTM-8699 in V63004" and another line mentioning the exact line item being tested i.e. "Previously the [PREVJNLLINKCUT error] message could be reported in rare cases where a journal switch was deferred, then later switched without cutting links"

JakeYottaDB: The appropriate comments have been added to lines 150 and 151

JakeYottaDB:
I've confirmed that the PREVJNLLINKCUT error is present in a v63003 run of the subtest

PR #248, opened 2018-06-20
 New v63003/gtm8847 subtest (Tests GTM-8847 in V63003)

VinayYottaDB:
The read/write (non-NEWable) $ZSTRPLLIM ISV provides a way for a process to limit its process private memory used for local variable and scratch storage. When the value is zero (0), the default, or negative, there is no limit. A positive value specifies a byte limit. When a request for additional memory exceeds the limit, GT.M does the expansion, and then produces an STPCRIT error. By default, a later request for memory produces an STPOFLOW, unless, subsequent to STPCRIT, $ZSTRPLLIM has been set to the same or higher limit. Note that GT.M allocates memory in large blocks so the interaction of $ZSTRPLLIM with memory growth is not exact. When the gtm_string_pool_limit environment variable specifies a positive value, GT.M uses it for the initial value of $ZSTRPLLIM. Previously, process memory was only limited by operating system configuration. (GTM-8847)


v63003/u_inref/gtm8847.csh
```diff
@@ -0,0 +1,29 @@
...
+#	under a license.  If you do not know the terms of	#
+#	the license, please stop and do not read further.	#
+#								#
+#################################################################
+#
+#
+#
+echo "# Attempting to overflow the string pool"
+$ydb_dist/mumps -run test2^gtm8847 >>& errors.outx
+set nonomatch;rm core*
```
nars1: Add a comment that "rm core*" is needed to remove corefile created by STPOFLOW error. Why is the nonomatch needed?

v63003/u_inref/gtm8847.csh
```diff
@@ -0,0 +1,29 @@
...
+#	This source code contains the intellectual property	#
+#	of its copyright holder(s), and is made available	#
+#	under a license.  If you do not know the terms of	#
+#	the license, please stop and do not read further.	#
+#								#
+#################################################################
+#
+#
+#
+echo "# Attempting to overflow the string pool"
```
nars1: Add a comment that we expect STPCRIT error first and then a STPOFLOW error

PR #256, opened 2018-06-25
Fix occasional mupjnl/ztp_tp_multi_reg subtest failure due to MUPIP JOURNAL region processing order being dependent on ftok order of *.dat files (not alphabetic order of *.dat files) and could change due to filesystem inode assignment of *.dat files

nars1:


mupjnl/u_inref/ztp_tp_multi_reg.csh
```diff
@@ -32,7 +35,11 @@ endif
 echo "----------------------------------------------------------------------------------"
 cp -f ./save/* .
 echo mupip journal -recover -back a.mjl,b.mjl -since=\"$time1\" -look=\"time=0 0:0:0\" -broken=a.broken
-$MUPIP journal -recover -back -since=\"$time1\" -look=\"time=0 0:0:0\" a.mjl,b.mjl -broken=a.broken
+# FILERENAME messages will appear below for each .mjl file that is renamed to .mjl_<timestamp>
+# But since the order in which MUPIP JOURNAL processes regions is the ftok order of the *.dat files
+# (not the alphabetic order), the output order of FILERENAME messages is not guaranteed to be a.mjl, b.mjl, c.mjl, d.mjl
+# Since the FILERENAME message is not pertinent to this test, just filter that non-deterministic part of the output away.
+$MUPIP journal -recover -back -since=\"$time1\" -look=\"time=0 0:0:0\" a.mjl,b.mjl -broken=a.broken |& $grep -v "FILERENAME"
```
estess: Question - which $status is tested here? The one for mupip or the one for grep?

nars1: Both. In a pipeline, if at least one command has a non-zero status, the final $status value will be non-zero. tcsh takes care of that.

PR #260, opened 2018-06-26
Fix gtm8699 addition to gtm8086 test

JakeYottaDB:
Previously gtm8699 code block c…ould pick up PREVJNLLINKCUT errors from other subtests in the syslog

### DIFF
--- a/v62000/u_inref/gtm8086.csh
+++ b/v62000/u_inref/gtm8086.csh
@@ -153,7 +153,7 @@ if ($?test_replic) then
 		# getoper will capture syslog messages generated since starting the jnlswitchretry error
 		$gtm_tst/com/getoper.csh "$jnlswitchretry_time"
 		# Confirm there is no PREVJNLLINKCUT error in the output
-		$grep -e "PREVJNLLINKCUT" syslog.txt
+		$grep -e "$PWD" syslog.txt | $grep -e "PREVJNLLINKCUT"
 
 
 	endif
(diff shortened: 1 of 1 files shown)


v62000/u_inref/gtm8086.csh
```diff
@@ -153,7 +153,7 @@ if ($?test_replic) then
 		# getoper will capture syslog messages generated since starting the jnlswitchretry error
 		$gtm_tst/com/getoper.csh "$jnlswitchretry_time"
 		# Confirm there is no PREVJNLLINKCUT error in the output
-		$grep -e "PREVJNLLINKCUT" syslog.txt
+		$grep -e "$PWD" syslog.txt | $grep -e "PREVJNLLINKCUT"
```
nars1: It is better to do the PREVJNLLINKCUT grep first as that will produce fewer lines for the second grep to process.
Also, why do we need the -e flag?

v62000/u_inref/gtm8086.csh
```diff
@@ -153,7 +153,7 @@ if ($?test_replic) then
 		# getoper will capture syslog messages generated since starting the jnlswitchretry error
 		$gtm_tst/com/getoper.csh "$jnlswitchretry_time"
 		# Confirm there is no PREVJNLLINKCUT error in the output
-		$grep -e "$PWD" syslog.txt | $grep -e "PREVJNLLINKCUT"
+		$grep "PREVJNLLINKCUT" | $grep "$PWD" syslog.txt
```
nars1: syslog.txt needs to be a parameter of the first grep.

PR #261, opened 2018-06-26
dbcreate.csh now supports -gld_has_db_fullpath option where .gld file points to full path of db (default is relative)

nars1:


VinayYottaDB:
Approved

PR #265, opened 2018-06-27
New r122/ydb297 subtest (Tests patch YDB297)

VinayYottaDB:

LOCK commands work correctly in rare cases when there are more than 31 subscripts at a given lock name subscript level (across all processes concurrently using a particular database file lock space) that hash to the same value. A $$^%PEEKBYNAME("sgmnt_data.lock_hash_bucket_full_cntr",REGION) now returns the # of times such a rare event was seen (and internally handled correctly) in the region named REGION. Previously, in this situation, it was possible for LOCK commands to spin-loop and incorrectly seize ownership of the lock resource from a concurrently holding process. This issue was introduced in GT.M V6.3-003 as part of GTM-8680 and so exists in YottaDB r1.20 and r1.22.


VinayYottaDB:
Not sure about the format of this type of subtest, if you could send me an example I can edit it when I do the squash and merge

nars1:
@VinayYottaDB : I will take care of the squash and merge this time. Thanks.

v63003/u_inref/ydb297.csh
```diff
@@ -0,0 +1,35 @@
...
+#								#
+#################################################################
+#
+# Tests that when over 31 locks hash to the same value, LOCK commands work correctly
+#
+$gtm_tst/com/dbcreate.csh mumps 1 >>& dbcreate.out
+# Simulates 2 errors in prior versions, either inappropriately seizes ownership of a lock or test hangs
+# Randomly chooses an option (1 will cause the test to hang, 0 will cause locks to be inappropriately seized)
+setenv hang `$gtm_tst/com/genrandnumbers.csh 1 0 1`
+$ydb_dist/mumps -run ydb297^ydb297 >>& job.out
```
nars1: You will need to add a "unsetenv ydb_lockhash_n_bits" before this line. This will make sure the test framework (which randomizes that env var) and code do not artificially cause further hash collisions than this test expects (which would cause test failures).

v63003/u_inref/ydb297.csh
```diff
@@ -0,0 +1,35 @@
...
+$gtm_tst/com/dbcheck.csh >>& dbcheck.out
+echo "# LOCKING RESOURCES THAT HASH TO THE SAME VALUE, EXPECT NUMBER OF LOCKS TO REMAIN CONSTANT ACROSS EACH RUN"
+echo "# INFO FROM RUN 1"
```
nars1: Mention here that "we don't expect to see any FAILURE string here". And do just the grep. No need to pipe the output to cat and get the result into $status etc. Likewise with RUN2 and 3.

v63003/u_inref/ydb297.csh
```diff
@@ -0,0 +1,35 @@
...
+set status1 = `cat lockhang_ydb297.mjo1 |& $grep FAILURE`
+cat lockhang_ydb297.mjo1 |& $grep LOCKSPACEINFO
+echo "# INFO FROM RUN 2"
+cat lockhang_ydb297.mjo2 |& $grep LOCKSPACEINFO
+set status2 = `cat lockhang_ydb297.mjo2 |& $grep SUCCESS`
+echo "# INFO FROM RUN 3"
+cat lockhang_ydb297.mjo3 |& $grep LOCKSPACEINFO
+set status3 = `cat lockhang_ydb297.mjo3 |& $grep SUCCESS`
+cat lockhang_ydb297.mjo3 |& $grep SEGMENT
+if (0 == `expr $hang` && "" == "$status1" && "" != "$status2" && "" != "$status3") then
```
nars1: Why do we need `expr $hang`. Why not just $hang. Based on prior comments, lines 33-35 might not be needed at all. Also notice an issue with such compound boolean expressions is that we will signal an inappropriate lock only if both status2 and status3 are non-zero whereas we should signal a failure as long as at least one of status2 and status3 are non-zero. Further confirms it is better to do the checks at each stage instead of combining all in one if check at the end.

v63003/u_inref/ydb297.csh
```diff
@@ -0,0 +1,35 @@
...
+#
+$gtm_tst/com/dbcreate.csh mumps 1 >>& dbcreate.out
+# Simulates 2 errors in prior versions, either inappropriately seizes ownership of a lock or test hangs
+# Randomly chooses an option (1 will cause the test to hang, 0 will cause locks to be inappropriately seized)
+setenv hang `$gtm_tst/com/genrandnumbers.csh 1 0 1`
+$ydb_dist/mumps -run ydb297^ydb297 >>& job.out
+$gtm_tst/com/dbcheck.csh >>& dbcheck.out
+echo "# LOCKING RESOURCES THAT HASH TO THE SAME VALUE, EXPECT NUMBER OF LOCKS TO REMAIN CONSTANT ACROSS EACH RUN"
+echo "# INFO FROM RUN 1"
+set status1 = `cat lockhang_ydb297.mjo1 |& $grep FAILURE`
```
nars1: cat | grep can be replaced with just grep. Avoids a pipeline.

PR #284, opened 2018-07-03
Fix gtm8889

JakeYottaDB:
change shell prompt to 'SHELL' in gtm8889.exp to avoid false positives from expect ">" commands. 
Avoid issue with inconsistent number of blank lines by grepping for non-blank lines in output in gtm8889.csh

### DIFF
--- a/v63003/outref/gtm8889.txt
+++ b/v63003/outref/gtm8889.txt
@@ -11,16 +11,16 @@ SHELLYDB
 $ydb_dist/mumps -dir
 YDB>zhelp
 Additional information available: 
-About_GT.M  Commands    Err_Processing          Functions   Integrate_External
-Internationalization    IO_Processing           ISV         Language_Extensions
-M_Lang_Features         Opr_Dbg_Dir_Mode        Program_Cycle           
-Triggers    Utility_Routines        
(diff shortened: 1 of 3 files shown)


v63003/outref/gtm8889.txt
```diff
@@ -4,33 +4,23 @@
...
-
-
 Additional information available: 
-
-About_YottaDB           Commands    Err_Processing          Functions
-Integrate_External      Internationalization    IO_Processing           
-ISV         Language_Extensions     M_Lang_Features         Opr_Dbg_Dir_Mode
-Program_Cycle           Triggers    Utility_Routines        
-
+About_GT.M  Commands    Err_Processing          Functions   Integrate_External
```
nars1: Why has this output changed? For example, About_YottaDB changed to About_GT.M

JakeYottaDB: This change was a product of running the gtm8889 test with the latest version v999_r122. If this was meant to run with a different version, I can do that and change the outref file accordingly.

nars1:
@JakeYottaDB : As discussed, this failure will take a lot more time to investigate and it happened once in 120 runs (very very rare) and it is not that critical an issue so we will go ahead with what we have now.

PR #288, opened 2018-07-05
Fix occasional unicode_io/unicodefifo subtest failure on slow systems

nars1:
In a test failure in-house, Line 38 below printed a FIFO reader couldn't start message
causing the failure. We started a job in line 36 and then wait 60 seconds for it to
execute 4 lines (starting from line 103) and set ^A=1 in line 107. But it did not.
Not sure exactly how long it took to reach line 107 so the timeout is now bumped to
 15 minutes (instead of 1 minute) assuming that should be more than good enough.
```
unicode_io/inref/unicodefifo.m
--------------------------------
   35    write jobstr,!
   36    job @jobstr
   37    for i=1:1:60 quit:^A  hang 1
   38    if 0=^A write "FIFO reader couldn't start, exiting",! quit
    .
    .
  102 unicodefifo2(file,encoding) ;
  103    set $ZTRAP="do error2^unicodefifo"
  104    write "$J:",$J,!
  105    write "file=",file,!
  106    write "encoding=",encoding,!
  107    set ^A=1
```

### DIFF
--- a/unicode_io/inref/unicodefifo.m
+++ b/unicode_io/inref/unicodefifo.m
@@ -3,6 +3,9 @@
 ; Copyright (c) 2006-2015 Fidelity National Information		;
 ; Services, Inc. and/or its subsidiaries. All rights reserved.	;
 ;								;
+; Copyright (c) 2018 YottaDB LLC. and/or its subsidiaries.	;
+; All rights reserved.						;
+;								;
 ;	This source code contains the intellectual property	;
 ;	of its copyright holder(s), and is made available	;
 ;	under a license.  If you do not know the terms of	;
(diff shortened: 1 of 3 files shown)


unicode_io/inref/unicodefifo.m
```diff
@@ -29,12 +32,12 @@
 	set ^A=1
 	do test(file,"FIFO:WRITE:RECORDSIZE=-1","",FAIL)
 	set ^A=0
-	f i=1:1:60 quit:^A  hang 1
+	f i=1:1:900 quit:^A  hang 1
```
estess: There are 8 time loops being modified in this routine as well as the other. Only the second one exits (the 3rd exits regardless) if the loop runs too long. It that worth doing for the others so we aren't turning an 8 minute max test into a 2 hr or more max test?

PR #290, opened 2018-07-06
 New v63002/gtm8644 subtest (Tests GTM-8644 in V63002)

VinayYottaDB:

The ZSYSTEM command always invokes shell specified by the SHELL environment variable; previously it was possible to construct arguments that did not. In addition, ZSYSTEM arguments handle quoting (both single and double) in a more straightforward way. This may require changes to existing code to simplify quoting. An example where a change may be needed:

GTM> zsystem "echo '""hello world""'"
hello world

will now produce:

GTM> zsystem "echo '""hello world""'"
"hello world"

and should be changed to:

GTM> zsystem "echo 'hello world'"
hello world

to get the same output. (GTM-8644)


v63002/u_inref/gtm8644.csh
```diff
@@ -0,0 +1,18 @@
...
+#	under a license.  If you do not know the terms of	#
+#	the license, please stop and do not read further.	#
+#								#
+#################################################################
+#
+#
+#
+setenv SHELL /usr/local/bin/tcsh
+$ydb_dist/mumps -run shellfn^gtm8644
+$ydb_dist/mumps -run quotesfn^gtm8644
```
nars1: I would additionally like a test of the fact that only $SHELL shows up instead of /bin/sh like the release note describes. Towards that, ps --forest -ef will give a good listing of the process tree while inside a ZSYSTEM and we expect to see only $SHELL there, no "sh".

PR #291, opened 2018-07-06
New v63002/gtm5250 subtest (Tests GTM-5250 in V63002)

VinayYottaDB:

For JOB, LOCK, OPEN and READ commands, and $gtm_tpnotacidtime, GT.M supports fractional timeout values to millisecond precision with timeout values capped at 999,999.999 seconds (about 11.5 days); higher timeout values execute with the capped value. Previously GT.M supported timeouts as integers (GTM-5250)


v63002/inref/gtm5250.m
```diff
@@ -0,0 +1,47 @@
...
+	tcommit
+	quit
+
+locktimeout
+	set jmaxwait=0
+	set ^X(1)=0
+	set ^X(2)=0
+	do ^job("lockchild^gtm5250",1,"""""")
+	for  quit:^X(1)  hang 1
+	lock ^A(1):1.234
```
nars1: I would like a test that a lock with timeout of 1.234 seconds times out somewhere between 1.23 and 1.24 seconds. That would be a test that fractional timeouts are really supported. One might not see that happening right away due to system load issues so one might need to keep trying this again say a max of 10 times before issuing an error. Same with OPEN command.

nars1: Forgot to mention. You can use $ZUT before and after the command and measure the difference and compare that value (in microseconds) against the timeout. Also, please randomly choose 1.2345 or 1.23456 instead of 1.234 to confirm that even if a precision more than millisecond is specified, it is rounded to the nearest millisecond.

v63002/inref/gtm5250.m
```diff
@@ -25,9 +25,14 @@
...
-	zwrite $test
-	write:'$test "Lock successfully timed out",!
+	for i=1:1:10 do  quit:(diff>1.234)&(diff<1.434) ;Giving a .2 second window for the timeout to occur, doing the same for open timeout
+	. set prev=$zut
+	. lock ^A(1):1.23456 ;Also testing that fractional timeouts specified to more than 3 decimal places get truncated
+	. set dollartest=$test
+	. set cur=$zut
+	. set diff=(cur-prev)*(10**-6)
+	write "$TEST=",dollartest,!
+	write:(diff>1.234)&(diff<1.434) "Lock successfully timed out",!
```
nars1: Also print failure in case of else condition

PR #292, opened 2018-07-06
New v63004/gtm8777 subtest (tests GTM-8777 in V63004)

JakeYottaDB:

When invoked at their QUIET or QCALL entrypoints, the %GCE, %GSE, %RCE, and %RSE utility routines report only globals or routines in which they find a match, for example: do QUIET^%GCE or do QCALL^%RSE. The QCALL entryref only exists in %RCE and %RSE. Previously these routines always reported every item they processed regardless of whether it contained a match. (GTM-8777)


v63004/u_inref/gtm8777.exp
```diff
@@ -0,0 +1,207 @@
...
+	timeout { timeout_procedure }
+}
+
+proc timeout_procedure { } {
+	puts "timeout occurred"
+	exit -1
+}
+
+#Start YDB
+send -- "\$gtm_dist/ydb\r"
```
nars1: ydb -> mumps -direct

This way one can test this against GT.M versions too (where ydb script does not exist).

v63004/u_inref/gtm8777.exp
```diff
@@ -0,0 +1,207 @@
...
+expect -exact "YDB>"
```
nars1: Same as above comment. Set YDB> as the prompt (using gtm_prompt env var). This will let us test this against a GT.M version.

v63004/u_inref/gtm8777.exp
```diff
@@ -0,0 +1,207 @@
...
+send -- "\$gtm_dist/mumps -direct\r"
+expect -exact "YDB>"
+
+#Start GCE prompt
+puts "QUIET^%GCE TEST\n"
+puts "--------------------------------------------------------------\n"
+send -- "do QUIET^%GCE\r"
+expect -exact "Global Change Every occurrence"
+expect -exact "Global ^"
+send -- "string1\r"
```
nars1: Can you do a ZWRITE ^string1 before and after the GCE call to confirm GCE did its job. Same case with ZWRITE ^string4 in next test. And with "cat test1.m" and 'cat test2.m" in the RCE tests.

PR #293, opened 2018-07-06
New v63002/gtm8760 subtest (Tests GTM-8760 in V63002)

VinayYottaDB:

GT.M now properly handles environment variables whose contents is over 32K in size. GT.M reports an error if the content is larger than can fit in a local or global. Previously, environment variables whose contents was over 32K in size resulted in a SIG-11. This issue was only observed in the GT.M development environment,and was never reported by a user. (GTM-8760)


v63002/inref/gtm8760.m
```diff
@@ -0,0 +1,38 @@
...
+;	under a license.  If you do not know the terms of	;
+;	the license, please stop and do not read further.	;
+;								;
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;
+longstring
+	set x="longstring.txt"
+	open x
+	use x
+	for i=1:1:$zcmdlne do
```
nars1: zcmdlne -> zcmdline

Also use the following line instead of lines 17-18. This is a lot faster.
```
write $translate($justify("a",+$zcmdline)," ","a")
```

v63002/u_inref/gtm8760.csh
```diff
@@ -0,0 +1,34 @@
...
+#	under a license.  If you do not know the terms of	#
+#	the license, please stop and do not read further.	#
+#								#
+#################################################################
+#
+# Testing a long string as a subscript and a value for global and local variables
+# All of these would produce core files in previous versions when the string is at least 32768 characters
+#
+$gtm_tst/com/dbcreate.csh mumps 1 >>& dbcreate.out
+foreach n (32766 32767 32768 32769)
```
nars1: Add 1048576 and 1048577 (i.e. 1Mb and 1Mb+1) to test that upto 1Mb works fine except as a subscript in globals. You will need to set the record_size to 1Mb (-record_size parameter to dbcreate.csh).

nars1: For the record, we tried the 1Mb suggestion but we ran into issues with the shell "Arg list too long" errors when running anything afterwards. So abandoned it.

PR #297, opened 2018-07-09
New v63004/gtm8916 subtest (tests GTM-8916 in V63004)

JakeYottaDB:
[PATCH SEEMS TO STILL SHOW SOME OF THE ERRORS DESCRIBED]

The receiver process recovers after its update process dies. Previously, on Linux systems with glibc 2.25 or newer, the receiver process could hang indefinitely, requiring manual cleanup of the process and shared memory. This issue was only observed in the GT.M development environment, and was never reported by a user. (GTM-8916)


JakeYottaDB:
The initial patch note says this is subtest gtm8699, which is incorrect. This is in fact, gtm8916.

v63004/outref/gtm8916.txt
```diff
@@ -0,0 +1,83 @@
+# Create a single region DB with region DEFAULT
+# Start replication
+==Executing MULTISITE_REPLIC 'START INST1 INST2'==
+Starting Primary Source Server in ##TEST_PATH##
+Starting Passive Source Server and Receiver Server in ##FILTERED##_REMOTE_TEST_PATH_/instance2
+# Start loop
+	# Iteration 1\:
+##TEST_AWK^==Executing MULTISITE_REPLIC 'RUN INST2 kill -9 .*
+==Executing MULTISITE_REPLIC 'RUN INST2 ##SOURCE_PATH##/mupip REPLIC -RECEIVE -START -UPDATEONLY # Restart update process'==
+%GTM-I-REPLINFO, Update Process started successfully
```
nars1: GTM -> YDB

Here and in various other places.

nars1:
@JakeYottaDB : As to your comment about this test still showing the hang even with the V63004 code fixes, our internal testing has confirmed that all boxes with glibc > 2.25 show the hang whereas they run the test fine when glibc <= 2.25. So this is likely a glibc bug/regression that will hopefully be fixed soon. Until then we will disable this subtest on boxes with glibc > 2.25.

v63004/u_inref/gtm8916.csh
```diff
@@ -0,0 +1,61 @@
...
+# All rights reserved.						#
+#								#
+#	This source code contains the intellectual property	#
+#	of its copyright holder(s), and is made available	#
+#	under a license.  If you do not know the terms of	#
+#	the license, please stop and do not read further.	#
+#								#
+#################################################################
+#
+
```
nars1: Need to do the following. Or else we could have a frozen instance (due to fake enospc) triggered by an update process which will never be unfrozen because it was kill -9ed by the test.

unsetenv gtm_test_freeze_on_error

PR #299, opened 2018-07-11
Fix rare v62001/gtm8094pipproc subtest failures due to missing syslog message

nars1:
We saw a failure in internal testing. Below is the actual output.
```
Expected output
---------------
Case 1: Prevent dup2() of stdin in forked piped process
Message Found!

Actual output
-------------
Case 1: Prevent dup2() of stdin in forked piped process
Message Not Found!
```
The syslog message was not seen in the Case 1.
```
$ cat msgs1.txt
-- Logs begin at Wed 2017-10-25 14:32:59 EDT, end at Tue 2018-07-10 16:59:22 EDT. --
-- No entries --
```
But it was seen in the syslog_gtm8094pipproc.txt file that is created at the end of the subtest.
```
$ cat syslog_gtm8094pipproc.txt
-- Logs begin at Wed 2017-10-25 14:32:59 EDT, end at Tue 2018-07-10 16:59:24 EDT. --
Jul 10 16:59:23 xxxx YDB-MUMPS[13825]: %YDB-E-DEVOPENFAIL, Error opening uname,
	 %YDB-I-TEXT, PIPE - dup2(pfd_write[0]) failed in child,
	 %SYSTEM-E-UNKNOWN, Unknown system error 0 -- generated from 0x00007F0161CB7122.
```
The time of delivery of the syslog message in syslog_gtm8094pipproc.txt is 16:59:23 whereas the
end time in msgs1.txt is 16:59:22. This means the syslog message was on its way to being
delivered when the getoper.csh call (that created msgs1.txt) was done but made it 1 second later.

To avoid such in-transit issues, we can use getoper.csh with $2 == "" which will cause getoper.csh
to send a syslog message (with a random string in it) and wait for that to show up which would
guarantee all previous messages have shown up too.

Additionally, to avoid multiple getoper.csh calls, it is done once at the end of the script. That
is okay since the syslog message corresponding to each test case has a unique message text.

### DIFF
--- a/v62001/outref/gtm8094pipproc.txt
+++ b/v62001/outref/gtm8094pipproc.txt
@@ -1,20 +1,16 @@
 Case 1: Prevent dup2() of stdin in forked piped process
 
-Message Found!
-
 Case 2: Prevent dup2() of stdout in forked piped process
-
-Message Found!
 
 Case 3: Prevent dup2() of stderr in forked piped process
 
-Message Found!
-
 Case 4: Prevent second dup2() of stderr in forked piped process
-
-
-Message Found!
 
 Case 5: Prevent EXEC() by the forked piped process
 
-Message Found!
+# See if there was one PIPE message in syslog for each Case above. List of messages follows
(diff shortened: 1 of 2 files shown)


nars1:
The test enhancements done in this pull request exposed a code issue which is fixed at YottaDB/YottaDB#310

PR #303, opened 2018-07-11
New v63002/gtm8717 subtest (Tests GTM-8717 in V63002)

VinayYottaDB:

$SELECT() appropriately handles an argument where a truthvalue made up of literals precedes a syntax error, as well as arguments where a literal FALSE truthvalue conditionalizes a global reference which a later argument also references. Due to a compiler optimization in V6.3-001[A], such an error produced a segmentation violation (SIG-11), and the other case produced a GTMASSERT2. (GTM-8717)


v63002/outref/gtm8717.txt
```diff
@@ -0,0 +1,6 @@
+# Testing 1000 random variations of $select syntax, checking for FATAL errors
```
nars1: Add a line saying we have seen some of these random variations generate GTMASSERT2 and KILLBYSIGSINFO1 fatal errors using V63001A

PR #308, opened 2018-07-13
New v63002/gtm8740 subtest (Tests GTM-8740 in V63002)

VinayYottaDB:

The command MUPIP REPLIC -SOURCE -JNLPOOL -SHOW displays whether the custom errors file is loaded. Previously the only mechanism to check this was via ^%PEEKBYNAME (available since V6.3-000) or $ZPEEK(). The following returns 0 when Instance Freeze is not enabled and 1 when it is: $$^%PEEKBYNAME("jnlpool_ctl_struct.instfreeze_environ_inited") When no prior custom errors file has been previously loaded, a Source Server starting with the gtm_custom_errors environment variable pointing to a file loads the contents of the file and thus enables Instance Freeze without performing a complete shutdown. Changing a previously loaded set of custom errors requires starting from a complete shutdown state. Previously if the Source Server initializing the replication journal pool was started without the gtm_custom_errors environment variable defined, enabling Instance Freeze by loading custom errors required a restart from a complete shutdown state. (GTM-8740)


v63002/u_inref/gtm8740.csh
```diff
@@ -0,0 +1,48 @@
...
+unsetenv gtm_custom_errors
+$MULTISITE_REPLIC_PREPARE 2
+$gtm_tst/com/dbcreate.csh mumps 1 >>& dbcreate.out
+
+$MSR START INST1 INST2
+echo ""
+$MUPIP REPLIC -SOURCE -JNLPOOL -SHOW >& replicshow.out
+$grep Custom replicshow.out
+$ydb_dist/mumps -run gtm8740
+# Starting a background process so the jnl pool does not get shut down
```
nars1: shut down -> shut down when the source server is stopped

nars1: Also need this line to be echoed in the reference file

v63002/u_inref/gtm8740.csh
```diff
@@ -0,0 +1,48 @@
...
+$MSR STOPSRC INST1 INST2
+echo ""
+echo "# Uploading custom errors"
+setenv gtm_custom_errors /dev/null
+echo ""
+
+echo "# Restarting primary"
+$MSR STARTSRC INST1 INST2
+# Ending background process
+$ydb_dist/mumps -run ^%XCMD "set ^stop=1"
```
nars1: Setting ^stop=1 is not enough. Need to also ensure the background process is dead before returning from this script.

v63002/u_inref/gtm8740.csh
```diff
@@ -0,0 +1,48 @@
...
+$MSR START INST1 INST2
+echo ""
+$MUPIP REPLIC -SOURCE -JNLPOOL -SHOW >& replicshow.out
+$grep Custom replicshow.out
+$ydb_dist/mumps -run gtm8740
+# Starting a background process so the jnl pool does not get shut down
+(($ydb_dist/mumps -run waitfn^gtm8740)&) > background.out
+echo ""
+echo "# Shutting down primary"
+$MSR STOPSRC INST1 INST2
```
nars1: Lines 17-18 can be moved to just before line 32. Line 47 can be moved to just after line 32. This keeps the scope of that env var as minimal as it can get.

PR #310, opened 2018-07-16
New v63002/gtm8616 subtest (Tests GTM-8616 in V63002)

VinayYottaDB:

MUPIP RUNDOWN with no arguments logs a message in the syslog facility with information about its actions including the process ID, user ID, and its current working directory. This may be useful for investigating any case where RUNDOWN inappropriately removes a GT.M shared memory segment or semaphore (IPC). Previously RUNDOWN did not provide such information. (GTM-8616)


v63002/u_inref/gtm8616.csh
```diff
@@ -0,0 +1,26 @@
...
+$gtm_tst/com/dbcreate.csh mumps 1 >>& dbcreate.out
+set t = `date +"%b %e %H:%M:%S"`
+sleep 1
+# Using .outx because expecting a warning about an argumentless MUPIP RUNDOWN
+echo "# Running argumentless MUPIP RUNDOWN and checking syslog"
+echo ""
+$MUPIP RUNDOWN >& rundown.outx
+$gtm_tst/com/getoper.csh "$t" "" getoper.txt
+echo "# Syslog message:"
+$grep "process id" getoper.txt |& sed 's/.*%YDB-I-MURNDWNARGLESS/%YDB-I-MURNDWNARGLESS/'
```
nars1: I would like the pid and username verified too. Not just the path. For pid, start the process in the background (which means you need to wait for this to die at the end) and $! will give you the pid.

v63002/u_inref/gtm8616.csh
```diff
@@ -19,8 +19,22 @@ sleep 1
...
+$gtm_tst/com/wait_for_proc_to_die.csh $pid
 $gtm_tst/com/getoper.csh "$t" "" getoper.txt
 echo "# Syslog message:"
 $grep "process id" getoper.txt |& sed 's/.*%YDB-I-MURNDWNARGLESS/%YDB-I-MURNDWNARGLESS/'
+set getoperpid=`$grep "process id" getoper.txt |& $tst_awk '{print $14}'`
+set getoperuid=`$grep "process id" getoper.txt |& $tst_awk '{print $17}'`
+if ($uid == $getoperuid) then
+	echo "# User ID in syslog message correct"
+else
+	echo "# Incorrect User ID in syslog message"
```
nars1: Please add "Expected [$uid] but found [$getoperuid]" to the error. Likewise to the pid in line 38. Anytime an error/failure is signaled, it is better to give as much information on the error.

PR #312, opened 2018-07-16
Fix rare v60001/gtm7461 subtest hang due to a test timing issue

nars1:
In an in-house test run, we had a test hang. This was because the instance got frozen
(by the backgrounded "mupip journal -rollback -backward" process) due to the JNLFILEOPNERR error
AFTER the foreground test script had unfreezed the instance resulting in a later call to
mupip rundown hanging due to the instance freeze active at that point.

The test flow has now been fixed so we wait for the backgrounded rollback to finish before
getting the syslog messages. We now confirm the REPLINSTFROZEN and JNLFILEOPNERR messages are
seen and include the actual message in the reference file.

### DIFF
--- a/v60001/u_inref/gtm7461.csh
+++ b/v60001/u_inref/gtm7461.csh
@@ -72,28 +72,28 @@ echo "# remove journal files"
 
 $gtm_tst/com/backup_dbjnl.csh "" "*.mjl*" "mv"
 
-echo "# rollback"
+echo "# start MUPIP JOURNAL -ROLLBACK -BACKWARD which should trigger instance freeze"
 
 set syslog_before1 = `date +"%b %e %H:%M:%S"`
 echo "$syslog_before1" > syslog_before1.txt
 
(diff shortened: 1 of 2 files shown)


v60001/u_inref/gtm7461.csh
```diff
@@ -72,29 +72,29 @@ echo "# remove journal files"
...
 $gtm_tst/com/backup_dbjnl.csh "" "*.mjl*" "mv"
 
-echo "# rollback"
+echo "# start MUPIP JOURNAL -ROLLBACK -BACKWARD which should trigger instance freeze"
 
 set syslog_before1 = `date +"%b %e %H:%M:%S"`
 echo "$syslog_before1" > syslog_before1.txt
 
-sleep 2		# Managed to slip out of the getoper window somehow, so give it a couple seconds to be safe.
+sleep 1		# Managed to slip out of the getoper window somehow, so give it a couple seconds to be safe.
```
estess: Comment still says "a couple seconds"..

PR #314, opened 2018-07-16
New v63002/gtm8718 subtest (Tests GTM-8718 in V63002)

VinayYottaDB:

Attempting to set an invalid string into $ZROUTINES leaves the previous value of $ZROUTINES as it was. Previously, this could cause issues if auto-relink was in use, resulting in a segmentation violation (SIG-11). This issue was only observed in the GT.M development environment, and was never reported by a user. (GTM-8718)


v63002/u_inref/gtm8718.csh
```diff
@@ -0,0 +1,28 @@
...
+#
+#
+#
+setenv gtm_test_autorelink_always 1
+set zroutines1 = `$ydb_dist/mumps -run ^%XCMD 'write $zroutines'`
+echo '# Changing $zroutines to an invalid string'
+$ydb_dist/mumps -run gtm8718
+echo ""
+set zroutines2 = `$ydb_dist/mumps -run ^%XCMD 'write $zroutines'`
+if ("$zroutines1" == "$zroutines2") then
```
nars1: $zroutines1 and $zroutines2 were derived from two different mumps -run invocations i.e. different processes. Those cannot be compared. They need to be derived from the same process and compared.

v63002/u_inref/gtm8718.csh
```diff
@@ -0,0 +1,28 @@
...
+#	This source code contains the intellectual property	#
+#	of its copyright holder(s), and is made available	#
+#	under a license.  If you do not know the terms of	#
+#	the license, please stop and do not read further.	#
+#								#
+#################################################################
+#
+#
+#
+setenv gtm_test_autorelink_always 1
```
nars1: What is the purpose of this line? Seems unnecessary.

VinayYottaDB: In the release note, it states if autorelink is in use, this ensures auto relink is in use

nars1: gtm_test_autorelink_always is a test system environment variable that is randomly chosen by the test framework at subtest startup. $zroutines is set with or without auto-relink based on the random choice before control reaches gtm8718.csh. So setting this env var does not help. You can try running this test with V63001A (which does not have the code fix) multiple times and you will see only 50% of the times it will core (i.e. when the random choice was auto-relink). It is okay for gtm8718.csh to run with the random choice as that improves test coverage and since the test output is the same in both cases.

nars1: As discussed, this line needs to be removed. Otherwise the changes are good to go.

PR #318, opened 2018-07-17
Fixed occasional v63002/gtm8766 test failure

VinayYottaDB:
We occasionally saw test failures that would fail because on a 32 bit box, the lower limit would be different than on a 64 big box. To circumvent this issue, we take note of what kind of box we are working with and set the lower limit accordingly, and modified the reference file to be more generalized to account for this

### DIFF
--- a/v63002/outref/gtm8766.txt
+++ b/v63002/outref/gtm8766.txt
@@ -1,5 +1,5 @@
-##TEST_AWK# In the current version, [0-9]* is the maximum value accepted by both MUPIP and GDE. In previous versions
-##TEST_AWK# [0-9]* was the max for MUPIP, but a core would be produced for a value greater than or equal to 2GB, and 2GB was the max value for GDE
+##TEST_AWK# In the current version, (2Mb|64Kb) is the maximum value accepted by both MUPIP and GDE. In previous versions
+##TEST_AWK# (2Mb|64Kb) was the max for MUPIP, but a core would be produced for a value greater than or equal to 2GB, and 2GB was the max value for GDE
 
 ##TEST_AWK# CHANGING THE BUFFER COUNT TO [0-9]*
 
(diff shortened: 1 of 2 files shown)


v63002/outref/gtm8766.txt
```diff
@@ -1,7 +1,7 @@
-# In the current version, 2MB is the maximum value accepted by both MUPIP and GDE. In previous versions
-# 2MB was the max for MUPIP, but a core would be produced for a value greater than or equal to 2GB, and 2GB was the max value for GDE
+##TEST_AWK# In the current version, [0-9]* is the maximum value accepted by both MUPIP and GDE. In previous versions
```
nars1: I would like (2Mb|64Kb) syntax instead of [0-9]*. That would ensure the only allowed choices are two instead of anything.

v63002/u_inref/gtm8766.csh
```diff
@@ -22,9 +22,17 @@ $GDE change -segment DEFAULT -access_method=BG >& accessmethod.out
 if ($status) then
 	echo "# UNABLE TO CHANGE ACCESS METHOD"
 endif
-echo "# In the current version, 2MB is the maximum value accepted by both MUPIP and GDE. In previous versions"
```
nars1: I would like this line reinserted but with a modification indicating 2Mb is for 64-bit builds and 64Kb is for 32-bit builds.

v63002/outref/gtm8766.txt
```diff
@@ -1,7 +1,7 @@
-# In the current version, 2MB is the maximum value accepted by both MUPIP and GDE. In previous versions
-# 2MB was the max for MUPIP, but a core would be produced for a value greater than or equal to 2GB, and 2GB was the max value for GDE
+##TEST_AWK# In the current version, (2Mb|64Kb) is the maximum value accepted by both MUPIP and GDE. In previous versions
```
nars1: (2Mb|64Kb) -> 2Mb-1 (for 64-bit builds) and 64Kb (for 32-bit builds)

v63002/u_inref/gtm8766.csh
```diff
@@ -22,9 +22,19 @@ $GDE change -segment DEFAULT -access_method=BG >& accessmethod.out
...
-echo "# 2MB was the max for MUPIP, but a core would be produced for a value greater than or equal to 2GB, and 2GB was the max value for GDE"
-foreach val (2097151 2097152 2147483647 2147483648)
+if ("64" == "$gtm_platform_size") then
+	@ smallvalminus1=2097151
+	@ smallval=2097152
+	set lowerlim="2Mb"
```
nars1: 2Mb -> 2Mb-1

PR #320, opened 2018-07-18
New v63002/gtm8711 subtest (Tests GTM-8711 in V63002)

VinayYottaDB:

When invoked from the shell, GDE appropriately maintains the return status; previously it did not. Also, the GT.M help facility handles an EOF on $PRINCIPAL; previously an EOF produced an internal error and generated an associated context dump file. (GTM-8711)


v63002/u_inref/gtm8711.csh
```diff
@@ -0,0 +1,21 @@
...
+#								#
+#	This source code contains the intellectual property	#
+#	of its copyright holder(s), and is made available	#
+#	under a license.  If you do not know the terms of	#
+#	the license, please stop and do not read further.	#
+#								#
+#################################################################
+#
+# Tests GDE appropriately maintains the return status when invoked from the shell
+#
```
nars1: The currently implemented test case invokes GDE and inside the GDE prompt sends an invalid command "asdfasdf". Would additionally like a test of "GDE asdfsdf" specified at the shell prompt and verify $status. Should be the same status as the first test.

v63002/u_inref/gtm8711.csh
```diff
@@ -0,0 +1,21 @@
...
+# All rights reserved.						#
+#								#
+#	This source code contains the intellectual property	#
+#	of its copyright holder(s), and is made available	#
+#	under a license.  If you do not know the terms of	#
+#	the license, please stop and do not read further.	#
+#								#
+#################################################################
+#
+# Tests GDE appropriately maintains the return status when invoked from the shell
```
nars1: Want to see the below release note also tested.

Also, the GT.M help facility handles an EOF on $PRINCIPAL; previously an EOF produced an internal error and generated an associated context dump file. (GTM-8711)

nars1: This comment is about the following release note.

*Also, the GT.M help facility handles an EOF on $PRINCIPAL; previously an EOF produced an internal error and generated an associated context dump file. (GTM-8711)*

Looks like this part of the change does not work as documented. At least when the terminal is the principal device, killing the terminal does produce a gtmhelpdmp.ZSHOW_DMP_* file in V63001A as well as V63002 so not sure what this release note means. So ignore my comment asking to test this change.

v63002/u_inref/gtm8711.csh
nars1: The below comment is the only one remaining to be addressed before I can approve this pull request.

*The currently implemented test case invokes GDE and inside the GDE prompt sends an invalid command "asdfasdf". Would additionally like a test of "GDE asdfsdf" specified at the shell prompt and verify $status. Should be the same status as the first test.*

v63002/outref/gtm8711.txt
```diff
@@ -1,21 +1,12 @@
...
-$ydb_dist/mumps -run ^GDE
-> $ydb_dist/mumps -run ^GDE
-%GDE-I-GDUSEDEFS, Using defaults for Global Directory 
-	##TEST_PATH##/mumps.gld
-
-GDE> # Sending Invalid Command, Expect The Shell To Maintain The Error Status
-asdfasdfasdf
-%GDE-E-KEYWRDBAD, ASDFASDFASDF is not a valid verb in this context
+# Running expect script in which we enter GDE and run an invalid command
+error status=114
```
nars1: I would also like to see the KEYWRDBAD error in this case in the reference file.

PR #334, opened 2018-07-24
 New v63002/gtm8733 subtest (Tests GTM-8733 in V63002)

VinayYottaDB:

$ZCONVERT() operates appropriately in UTF-8 NOBADCHAR mode; previously it issued an error. As a consequence the pinentry script operates correctly. The pinentry issue was only observed in the GT.M development environment, and was never reported by a user. Note that defining the gtm_ztrap_form environment variable to the empty (null) string sets ZTRAP handling to adaptive, while leaving it undefined sets the handling to "code" which can cause entryref device exception or ZTRAP handlers to fail with the error INDEXTRACHARs. (GTM-8733)


v63002/inref/gtm8733.m
```diff
@@ -0,0 +1,17 @@
...
+;	of its copyright holder(s), and is made available	;
+;	under a license.  If you do not know the terms of	;
+;	the license, please stop and do not read further.	;
+;								;
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;
+gtm8733
+	view "NOBADCHAR"
+	set z=$zconvert($zch(128),"U")
+	zwrite
```
nars1: Add below as a comment

; Note: Cannot use "write" as that would cause a BADCHAR error even with VIEW "NOBADCHAR"

v63002/outref/gtm8733.txt
```diff
@@ -0,0 +1,5 @@
+# Switching to UTF-8 Mode
+setenv gtm_chset UTF-8
+unset echo
+# Running $ZCONVERT in NOBADCHAR mode
```
nars1: mode -> mode on $ZCHAR(128) (which is an invalid UTF8 byte sequence) but we expect no BADCHAR error

PR #335, opened 2018-07-24
 New v63002/gtm8698 subtest (Tests GTM-8698 in V63002)

VinayYottaDB:

When GT.M encounters an error creating an autoDB database (including those used to store statistics), it issues a MUCREFILERR which identifies the application code entryref where it encountered the issue followed by the underlying error. Previously the it only supplied the underlying error with no context as to where it occurred. In addition, if GT.M encounters an error running down a statistics database, it reports the error to the syslog; previously reported to the user's device. (GTM-8698)


v63002/instream.csh
```diff
@@ -22,13 +22,14 @@
 # gtm8760	    [vinay] Tests YottaDB properly handles environment variables whose contesnts are over 32K in size
 # gtm5250	    [vinay] Tests YottaDB supports fractional timeouts
 # gtm8736	    [vinay] Tests $zroutines defaults to "." when $gtmroutines is undefined
+# gtm8698	    [vinay] Tests GTM issues a MUCREFILERR identifying the application code entryref when it encounters an error creating an AutoDB Database
```
nars1: GTM -> YottaDB
MUCREFILERR -> MUCREFILERR error in syslog

v63002/u_inref/gtm8698.csh
```diff
@@ -0,0 +1,34 @@
...
+#
+#
+$gtm_tst/com/dbcreate.csh mumps 1 >>& dbcreate.out
+echo "# Changing the region to AUTODB and setting the file to an invalid .dat file"
+echo "# This will trigger an error when we attempt to update the database"
+$GDE change -segment DEFAULT -file_name="a/a.dat" >& changefile.out
+$GDE change -region DEFAULT -AUTODB >& autodb.out
+rm mumps.dat
+set t = `date +"%b %e %H:%M:%S"`
+sleep 1
```
nars1: This sleep seems unnecessary.

v63002/u_inref/gtm8698.csh
```diff
@@ -0,0 +1,34 @@
...
+echo ""
+echo "# Attempting to update the database"
+$ydb_dist/mumps -run ^%XCMD 'set ^A=1'
+echo "# Checking the syslog"
+$gtm_tst/com/getoper.csh "$t" "" getoper.txt
+$grep "MUCREFILERR" getoper.txt |& sed 's/.*%YDB-E-MUCREFILERR/%YDB-E-MUCREFILERR/'|& sed 's/%YDB-E-DBOPNERR.*//'
+# Fixing everything so the dbcheck can operate appropriately
+$GDE change -segment DEFAULT -file_name="mumps.dat" >& fix.out
+$ydb_dist/mumps -run ^%XCMD 'set ^A=1' >& setvar.out
+$gtm_tst/com/leftover_ipc_cleanup_if_needed.csh $0
```
nars1: Why is this line needed?

v63002/u_inref/gtm8698.csh
VinayYottaDB: I believe the way getoper works is it starts checking the first second after the set time, meaning if the message were to come in the same second, it will not pick it up. As such, I have adopted the practice of always including the sleep 1 after I set the time to avoid this (I did not observe this behavior on this test but it has caused me issues on other tests)

nars1:
For the record. I tried to see how we can test the below release note but could not find any code changes or test case that demonstrates any change. So this is not being tested.

*In addition, if GT.M encounters an error running down a statistics database, it reports the error to the syslog; previously reported to the user's device.*

v63002/u_inref/gtm8698.csh
VinayYottaDB: It looks like it was not, I think i included it in an attempt to remedy some errors earlier on but it was actually unrelated

v63002/u_inref/gtm8698.csh
nars1: As discussed, getoper.csh gets messages from the start of the set time, not 1 second after. So we will remove any sleeps whose need we are not sure of. And reintroduce them (with comments on their need) if we later realize the need.

PR #339, opened 2018-07-25
New v63005/gtm8877 subtest (Tests GTM-8877 in V63005 and YottaDB/YottaDB#318)

VinayYottaDB:

The GT.M restriction mechanism recognizes the following lines:

ZSYSTEM_FILTER[:M labelref]
PIPE_FILTER[:M labelref]

The labelref must include a routine name. If a process is restricted by a ZSYSTEM or PIPE_OPEN line in the restrictions file that restriction takes precedence over the corresponding filter restriction. Otherwise when a process is subject to these restrictions, GT.M inserts an invocation of the labelref prior to the restricted command, passing a string containing the argument to the ZSYSTEM command or the command deviceparameter of the PIPE OPEN. The path to the filter routine must be included in $zroutines. FIS recommends that the filter routine is placed in a location with restricted access such as $gtm_dist. If the filter invocation return is -1,GT.M produces a RESTRICTEDOP error, otherwise it executes the command using the returned string via output parameters as a, possibly identical, replacement for the original string. Since GT.M uses the call-ins mechanism to execute the filters, a filter invocation inside a TP transaction in call-ins produces a CITPNESTED error. Note that because ZSYSTEM and OPEN are not Isolated actions FIS recommends against their use within a TP transaction. Filters also increment the nested level of call-ins. A recursive filter invocation produces a NOFILTERNEST error. GT.M reports all filter errors to the operator log accompanied by a COMMFILTERERR.

An example restrict file for this 

ZSYSTEM_FILTER:^filterzsy
PIPE_FILTER:^filterzsy

The actual filter routine:

```
filterzsy(inarg,outarg);

if ""=inarg set outarg="-1;must provide a command" quit

for i=1:1 set arg=$piece(inarg,";",i) quit:""=arg  do  quit:$data(outarg)

. for  quit:$zchar(9,32)'[$extract(arg)  set arg=$extract(arg,2,9999)

. set cmd=$piece(arg," ")

. for restrict="sudo","cd" if cmd=restrict set outarg="-1;command "_restrict_" not permitted" quit

. quit:$data(outarg)

. if "echo"=cmd set $piece(arg," ")="echo #",$piece(inarg,";",i)=arg    ;example of modification

set:'$data(outarg) outarg=inarg

quit +outarg
```

Filter execution starts with $STACK=1 ($ZLEVEL=2).

Following are the GT.M commands, Intrinsic Special Variables, and functions whose behavior changes in the context of a filter invocation.

ZGOTO 0 (zero) returns to the processing of the restricted command as does ZGOTO 1 (one) with no entryref, while ZGOTO 1:entryref replaces the originally invoked filter and continues filter execution.

$ZTRAP/$ETRAP NEW'd at level 1.

$ZLEVEL initializes to one (1) in GTM$CI, and increments for every new stack level.

$STACK initializes to zero (0) in GTM$CI frame, and increments for every new stack level.

$ESTACK NEW'd at level one (1) in GTM$CI frame.

$ECODE/$STACK() initialized to the empty string at level one (1) in GTM$CI frame.

After the filter completes, GT.M restores the above to their values at the invocation of the filter. (GTM-8877) 

Also tests YottaDB/YottaDB318:

Changes to ZSYSTEM_FILTER/PIPE_FILTER in restrict.txt (which resides in $ydb_dist/$gtm_dist) correctly take effect for future ZSYSTEM/PIPE OPEN commands. In GT.M V6.3-005, where this feature was introduced, if restrict.txt is changed multiple times within the same second for ZSYSTEM_FILTER/PIPE_FILTER, the changes did not take effect. A workaround was to "touch" the restrict.txt file afterwards to make its timestamp newer (relative to an auto-generated file filter_commands.tab).


v63005/u_inref/gtm8877.csh
```diff
@@ -0,0 +1,86 @@
...
+
+echo "# NO RESTRICTIONS"
+$ydb_dist/mumps -run zsystemfn^gtm8877
+$ydb_dist/mumps -run pipeopenfn^gtm8877
+echo ""
+
+
+# FILTER RETURNING -1
+cat > $ydb_dist/restrict.txt << EOF
+ZSYSTEM_FILTER:filterfn^gtm8877
```
nars1: Can we have a test where ZSYSTEM_FILTER and PIPE_FILTER are specified but without the : or entryref. To see what happens.

v63005/outref/gtm8877.txt
```diff
@@ -0,0 +1,30 @@
+# NO RESTRICTIONS
+original input
+original input
+
+# FILTER RETURNING -1
+%YDB-E-COMMFILTERERR, Error executing the command filter for gtmzsystem. RESTRICTEDOP
+%YDB-E-RESTRICTEDOP, Attempt to perform a restricted operation: echo 'original input'
+%YDB-E-COMMFILTERERR, Error executing the command filter for gtmpipeopen. RESTRICTEDOP
+%YDB-E-RESTRICTEDOP, Attempt to perform a restricted operation: echo 'original input'
+ZSTATUS=pipeopenfn+4^gtm8877,%YDB-E-IONOTOPEN, Attempt to USE an I/O device which has not been opened
```
nars1: Why do we see a IONOTOPEN error?

v63005/outref/gtm8877.txt
```diff
@@ -0,0 +1,30 @@
...
+
+# FILTER RETURNING A STRING
+filtered output
+filtered output
+
+# FILTER INVOCATION IN A TP TRANSACTION
```
nars1: Can we have an additional test of FILTER INVOCATION IN A TP TRANSACTION WITH RESTARTS (the restarts can be induced by a zsystem $incr just like was done in v63002/gtm8165 for the tpnotacid case.

v63005/outref/gtm8877.txt
```diff
@@ -0,0 +1,30 @@
...
+# FILTER WITH ZSYSTEM AND PIPE_OPEN RESTRICTED
+ZSTATUS=zsystemfn+2^gtm8877,%YDB-E-RESTRICTEDOP, Attempt to perform a restricted operation: ZSYSTEM
+ZSTATUS=pipeopenfn+3^gtm8877,%YDB-E-RESTRICTEDOP, Attempt to perform a restricted operation: OPEN PIPE
+
+# RECURSIVE FILTERS
+ZSTATUS=reczsystfilter+2^gtm8877,%YDB-E-NOFILTERNEST, Filter nesting not allowed
+%YDB-E-COMMFILTERERR, Error executing the command filter for gtmzsystem. 150374450,reczsystfilter+3^gtm8877,%YDB-E-NOTEXTRINSIC, QUIT/ZHALT does not return to an extrinsic function: argument not allowed
+ZSTATUS=recpipefilter+3^gtm8877,%YDB-E-NOFILTERNEST, Filter nesting not allowed
+%YDB-E-COMMFILTERERR, Error executing the command filter for gtmpipeopen. 150374450,recpipefilter+4^gtm8877,%YDB-E-NOTEXTRINSIC, QUIT/ZHALT does not return to an extrinsic function: argument not allowed
+ZSTATUS=pipeopenfn+4^gtm8877,%YDB-E-IONOTOPEN, Attempt to USE an I/O device which has not been opened
```
nars1: Same question. Why is this IONOTOPEN error happening?

v63005/outref/gtm8877.txt
```diff
@@ -0,0 +1,30 @@
...
+filtered output
+
+# FILTER INVOCATION IN A TP TRANSACTION
+filtered output
+filtered output
+
+# FILTER WITH ZSYSTEM AND PIPE_OPEN RESTRICTED
+ZSTATUS=zsystemfn+2^gtm8877,%YDB-E-RESTRICTEDOP, Attempt to perform a restricted operation: ZSYSTEM
+ZSTATUS=pipeopenfn+3^gtm8877,%YDB-E-RESTRICTEDOP, Attempt to perform a restricted operation: OPEN PIPE
+
```
nars1: I would also like a test for the below portion of the release note.

*GT.M reports all filter errors to the operator log accompanied by a COMMFILTERERR.*

v63005/outref/gtm8877.txt
VinayYottaDB: The opening of the pipe device is unssucessful due to the filter, and so in the next line it attempts to use a device that had not been opened

nars1: Then check $test to see if the open succeeded and only if it did, do the use. Otherwise do not do the use. That way we avoid these unrelated errors.

v63005/u_inref/gtm8877.csh
VinayYottaDB: System treats it as a syntax error and subsequently restricts everything (which happens when there is a syntax error in the restrict.txt). With this, the pecking order of ZSYSTEM and PIPE_OPEN vs ZSYSTEM_FILTER and PIPE_FILTER gets invoked, and it treats it as an invalid operation rather than a filtered operation

nars1: Thanks. I have made sure we fix the GTM-88877 release note (to indicate the M labelref is mandatory and not optional) when we release r1.24 with the GTM-8877 fixes.

v63005/outref/gtm8877.txt
nars1: As discussed, $test is not set in case OPEN fails and we handle the error with ^incrtrap. So we let the IONOTOPEN error stay in the reference file wherever it shows up.

com/SUITE
```diff
@@ -179,6 +179,8 @@ v63003	E
 v63003	E REPLIC
 v63004	E
 v63004	E REPLIC
+v63005  E
```
nars1: This change is already done.

v63005/outref/gtm8877.txt
```diff
@@ -28,6 +43,17 @@ ZSTATUS=recpipefilter+3^gtm8877,%YDB-E-NOFILTERNEST, Filter nesting not allowed
...
+# Checking the Sys Log
+##TEST_AWK%YDB-E-NOFILTERNEST, Filter nesting not allowed -- generated from 0x.*
+##TEST_AWK%YDB-E-NOFILTERNEST, Filter nesting not allowed -- generated from 0x.*
+##TEST_AWK%YDB-E-COMMFILTERERR, Error executing the command filter for gtmzsystem. 150374450,reczsystfilter\+3\^gtm8877,%YDB-E-NOTEXTRINSIC, QUIT/ZHALT does not return to an extrinsic function: argument not allowed -- generated from 0x.*
+##TEST_AWK%YDB-E-COMMFILTERERR, Error executing the command filter for gtmpipeopen. 150374450,recpipefilter\+4\^gtm8877,%YDB-E-NOTEXTRINSIC, QUIT/ZHALT does not return to an extrinsic function: argument not allowed -- generated from 0x.*
+
 # No label specified in filters
 ZSTATUS=zsystemfn+2^gtm8877,%YDB-E-RESTRICTEDOP, Attempt to perform a restricted operation: ZSYSTEM
 ZSTATUS=pipeopenfn+3^gtm8877,%YDB-E-RESTRICTEDOP, Attempt to perform a restricted operation: OPEN PIPE
+# Checking th Sys Log
```
nars1: Why are there 4 RESTRICTSYNTAX messages? We only do 2 mumps invocations. The reason is best placed in this reference file as a comment.

v63005/u_inref/gtm8877.csh
```diff
@@ -112,7 +112,9 @@ chmod -w $ydb_dist/restrict.txt
 echo "# No label specified in filters"
 $ydb_dist/mumps -run zsystemfn^gtm8877
 $ydb_dist/mumps -run pipeopenfn^gtm8877
-echo "# Checking th Sys Log"
+echo "# Checking the Sys Log"
+# getoper.csh calls an m script, which creates excess RESTRICTSYNTAX errors in the syslog
```
nars1: Reword as "Avoid extra RESTRICTSYNTAX errors from getoper.csh (which does mumps -run) by removing restrict.txt before the call"

v63005/outref/gtm8877.txt
```diff
@@ -52,8 +52,6 @@ ZSTATUS=pipeopenfn+4^gtm8877,%YDB-E-IONOTOPEN, Attempt to USE an I/O device whic
 # No label specified in filters
 ZSTATUS=zsystemfn+2^gtm8877,%YDB-E-RESTRICTEDOP, Attempt to perform a restricted operation: ZSYSTEM
 ZSTATUS=pipeopenfn+3^gtm8877,%YDB-E-RESTRICTEDOP, Attempt to perform a restricted operation: OPEN PIPE
-# Checking th Sys Log
-##TEST_AWK%YDB-E-RESTRICTSYNTAX, Syntax error in file ##TEST_PATH##/ydb_temp_dist/restrict.txt at line number 1. All facilities restricted for process. -- generated from 0x.*
-##TEST_AWK%YDB-E-RESTRICTSYNTAX, Syntax error in file ##TEST_PATH##/ydb_temp_dist/restrict.txt at line number 1. All facilities restricted for process. -- generated from 0x.*
+# Checking the Sys Log
```
nars1: Sys Log -> syslog

PR #340, opened 2018-07-25
New v63002/gtm6657 subtest (Tests GTM-6657 in V63002)

VinayYottaDB:

MUPIP BACKUP -BKUPDBJNL=OFF does not adjust the journaling state for a database with journaling disabled. Previously, -BKUPDBJNL=OFF turned off journaling in the backup database even when journaling was disabled, which required an extra step to explicitly reestablish the disabled journaling state characteristic for regions in the restored database that originally had journaling disabled. (GTM-6657)


v63002/inref/gtm6657.m
```diff
@@ -0,0 +1,13 @@
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
```
nars1: Empty file can be removed.

v63002/outref/gtm6657.txt
```diff
@@ -0,0 +1,10 @@
+# Disabling Journaling on the Database
+%YDB-I-JNLSTATE, Journaling state for region DEFAULT is now DISABLED
+# Backing up Database with -BKUPDBJNL=OFF
+%YDB-I-JNLSTATE, Journaling state for database file backup is now DISABLED
```
nars1: This DISABLED output is from the backup command. Additionally, I would like that we copy the backup database onto mumps.dat and do a DSE DUMP -FILE |& $grep "Journal State" to confirm that it is indeed DISABLED and not OFF.

v63002/u_inref/gtm6657.csh
```diff
@@ -18,5 +18,7 @@ echo "# Disabling Journaling on the Database"
 $MUPIP set -region default -journal=disable
 echo "# Backing up Database with -BKUPDBJNL=OFF"
 $MUPIP BACKUP -BKUPDBJNL=OFF default backup
+cp backup mumps.dat
```
nars1: Add echo"# Verify journal state in backup database using DSE DUMP -FILE"

PR #343, opened 2018-07-25
Test enhancements and fixes due to #312/GTM-8182 code fixes

nars1:


com/tpntpupd.m
```diff
@@ -165,14 +172,18 @@
 	if operation=5 do noop
 	quit
 
+tpnomultigldcheck(str)
+	if tpnomultigld set str="set $zgbldir=tpgld "_str
```
estess: Could be written as below to avoid the if logic (generates better code):

set:tpnomultigld str="..."...

No need to change this one or the several others in here but perhaps keep in mind for next time.

nars1: Addressed in https://github.com/YottaDB/YottaDBtest/pull/359

PR #347, opened 2018-07-26
New v63005/gtm8930 subtest (tests GTM-8930 in V63005)

JakeYottaDB:

If the process has access to an instance file designation, $VIEW("JNLPOOL") returns its name; if the process has not yet opened the pool, the return contains an asterisk after the name. If the process does not have sufficient information to determine a replication journal instance file, the function returns "No replication instance defined." Previously, the function with this argument returned an empty string if the process had not opened the journal pool. (GTM-8930)


v63005/inref/gtm8930.m
```diff
@@ -0,0 +1,36 @@
...
+gtm8930
+	quit
+
+test1
+	SET tab="    "
+
+	WRITE tab_"$VIEW(""JNLPOOL"") with unopened JNLPOOL",!
+	WRITE tab_$VIEW("JNLPOOL"),!!
+
+	WRITE tab_"Updatig DB",!!
```
nars1: Updatig -> Updating

v63005/u_inref/gtm8930.csh
```diff
@@ -0,0 +1,67 @@
...
+	cat dbcreate_log.txt
+endif
+echo ''
+
+echo "# Start replication"
+$MSR START INST1 INST2
+echo ''
+
+
+echo '# Run test1^gtm8930.m to test $VIEW("JNLPOOL") before and after opening JNLPOOL'
```
nars1: In this test, I want "gtm_repl_instance" randomly set to mumps.repl or to a value other than mumps.repl. To verify that the env var has no bearing on the jnlpool as long as the gld has the mapping defined.

v63005/u_inref/gtm8930.csh
```diff
@@ -0,0 +1,67 @@
...
+$ydb_dist/mumps -run test1^gtm8930
+echo ''
+
+echo "# Remove instance file mapping in GDE"
```
nars1: I want a test where instance file mapping in GDE is removed but gtm_repl_instance is set to an arbitrary file name. In that case, want to see $view("jnlpool") returns the value in the env var.

v63005/u_inref/gtm8930.csh
```diff
@@ -0,0 +1,67 @@
...
+$GDE CHANGE -INSTANCE -FILE_NAME=\"\" >& gdeChange.txt
+echo ''
+
+echo "# Unset gtm_repl_instance and ydb_repl_instance"
```
nars1: ydb_repl_instance usages in the test system can be removed. Throughout the test system we use the gtm_* variables as it avoids duplication of code. Only in certain tests do we explicitly test the ydb_* env vars.

v63005/u_inref/gtm8930.csh
```diff
@@ -0,0 +1,67 @@
...
+echo ''
+
+echo "# Re-set gtm_repl_instance and ydb_repl_instance for shutdown"
+setenv ydb_repl_instance "mumps.repl"
+setenv gtm_repl_instance $ydb_repl_instance
+echo "ydb_rep_instance: $ydb_repl_instance"
+echo "gtm_rep_instance: $gtm_repl_instance"
+echo ''
+
+echo "Stop replication"
```
nars1: No need for lines 57-59. dbcheck takes care of that.

v63005/inref/gtm8930.m
```diff
@@ -0,0 +1,36 @@
...
+	SET ^Tenth="David Tennant"
+
+	WRITE tab_"$VIEW(""JNLPOOL"") with opened JNLPOOL",!
+	WRITE tab_$VIEW("JNLPOOL"),!!
```
nars1: In line 26 and 34, make sure the output shows up as

$VIEW("JNLPOOL")=...

v63005/u_inref/gtm8930.csh
```diff
@@ -29,36 +30,49 @@ echo "# Start replication"
...
-$MSR STOP INST1 INST2
+echo '# Run test3^gtm8930.m to test $VIEW("JNLPOOL") with no instance file mapped and gtm_repl_instance set to an arbitrary value'
+setenv gtm_repl_instance "arbitrary.repl"
+$ydb_dist/mumps -run test3^gtm8930
 echo ''
 
 
+echo "# Re-set gtm_repl_instance for shutdown"
+setenv gtm_repl_instance "mumps.repl"
+echo "gtm_rep_instance: $gtm_repl_instance"
```
nars1: No need for this line. Line 71 is good enough.

PR #360, opened 2018-08-01
com/expectinit.exp Implemented in v63004 subtests gtm8777, gtm8791, and gtm8909

JakeYottaDB:
Add com/expectinit.exp tool to uniformly setup expect scripts in the test framework. Implement com/expectinit.exp in all v63004 expect scripts (gtm8777, gtm8791, gtm8909).

### DIFF
--- a/com/expectinit.exp
+++ b/com/expectinit.exp
@@ -1,0 +1,54 @@
+#################################################################
+#								#
+# Copyright (c) 2018 YottaDB LLC. and/or its subsidiaries.	#
+# All rights reserved.						#
+#								#
+#	This source code contains the intellectual property	#
+#	of its copyright holder(s), and is made available	#
+#	under a license.  If you do not know the terms of	#
+#	the license, please stop and do not read further.	#
+#								#
+#################################################################
+#
+# expectinit.exp is meant to be sourced from all test framework expect scripts in order
+# to set up each expect environment uniformly. This way test output does not depend upon
+# what the shell is like where the expect is called from.
+#
+#	USAGE:  source from within your expect script AFTER spawning your expect scripts shell
+#
+#		set timeout 60					#time out is set independently
+#		spawn /usr/local/bin/tcsh -f			#spawn shel first
+#		source $::env(gtm_tst)/com/expectinit.exp	#source expetinit.exp
+#
+# 	Note: environment variabls in .exp srcipts must be accessed with $::env(<var>)
+
+
+# Have columns higher than 80 as that can cause test failures on lines that are just above 80 columns in length
+expect -exact ">"
+send -- "stty cols 132\r"
+expect "stty cols 132\r"
+expect -exact ">"
+
+# Change shell prompt to something other than ">" as that is a substring of many YDB prompts
+# and can cause incorrect match later when we wait for the shell prompt.
+expect -exact ">"
+# Note: Changing the shell prompt to SHELL might seem easily achieved as follows.
+#	send -- "set prompt=SHELL\r"
+#	expect -exact "SHELL"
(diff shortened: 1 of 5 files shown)


com/expectinit.exp
```diff
@@ -0,0 +1,54 @@
...
+#################################################################
+#
+# expectinit.exp is meant to be sourced from all test framework expect scripts in order
+# to set up each expect environment uniformly. This way test output does not depend upon
+# what the shell is like where the expect is called from.
+#
+#	USAGE:  source from within your expect script AFTER spawning your expect scripts shell
+#
+#		set timeout 60					#time out is set independently
+#		spawn /usr/local/bin/tcsh -f			#spawn shel first
```
nars1: shel -> shell

com/expectinit.exp
```diff
@@ -0,0 +1,54 @@
...
+#		source $::env(gtm_tst)/com/expectinit.exp	#source expetinit.exp
```
nars1: expetinit -> expectinit

com/expectinit.exp
```diff
@@ -0,0 +1,54 @@
...
+#								#
+#	This source code contains the intellectual property	#
+#	of its copyright holder(s), and is made available	#
+#	under a license.  If you do not know the terms of	#
+#	the license, please stop and do not read further.	#
+#								#
+#################################################################
+#
+# expectinit.exp is meant to be sourced from all test framework expect scripts in order
+# to set up each expect environment uniformly. This way test output does not depend upon
```
nars1: Remove the second sentence beginning with "This way test output ..."

com/expectinit.exp
```diff
@@ -0,0 +1,54 @@
...
+#	expect -exact "SHELL"
+# But that will not work because it is possible the "expect" matches the SHELL from the "set prompt=SHELL" input.
+# To avoid this, we first store the "SHELL" string in a variable and use that to set the prompt.
+send -- "set shellprompt=SHELL\r"
+expect -exact ">"
+send -- "set prompt=\$shellprompt\r"
+expect -exact "SHELL"
+
+
+# Set uniform behaivor for handeling a timeout
```
nars1: handeling -> handling

com/expectinit.exp
```diff
@@ -0,0 +1,54 @@
...
+#	the license, please stop and do not read further.	#
+#								#
+#################################################################
+#
+# expectinit.exp is meant to be sourced from all test framework expect scripts in order
+# to set up each expect environment uniformly. This way test output does not depend upon
+# what the shell is like where the expect is called from.
+#
+#	USAGE:  source from within your expect script AFTER spawning your expect scripts shell
+#
```
nars1: Add a note here that the shell prompt is reset to SHELL and so that is what the parent script should be looking for if/when they expect the shell prompt.

com/expectinit.exp
```diff
@@ -0,0 +1,54 @@
...
+#		set timeout 60					#time out is set independently
+#		spawn /usr/local/bin/tcsh -f			#spawn shel first
+#		source $::env(gtm_tst)/com/expectinit.exp	#source expetinit.exp
+#
+# 	Note: environment variabls in .exp srcipts must be accessed with $::env(<var>)
+
+
+# Have columns higher than 80 as that can cause test failures on lines that are just above 80 columns in length
```
nars1: Replace this line with
```
# Have columns set to a high value of 132 as many callers have long lines
#   and we do not want line wrapping to cause test failures.
# Also note that the expect terminal width is inherited from the parent terminal
#  from where the test is invoked and so setting this to 132 avoids false failures
#  due to the parent terminal having an arbitrarily (small) terminal width.
```

JakeYottaDB:
Sorry about all of those typos. Even for me that was an excessive amount.

PR #363, opened 2018-08-06
Updated v63002/gtm8694 subtest 

VinayYottaDB:
Added a case to the subtest to test that a RESTRICTSYNTAX error is produced in the syslog when an m program is called and there is a syntax error in restrict.txt

### DIFF
--- a/v63002/u_inref/gtm8694.csh
+++ b/v63002/u_inref/gtm8694.csh
@@ -254,12 +254,11 @@ $gtm_tst/com/lsminusl.csh $ydb_dist/restrict.txt | $tst_awk '{print $1,$9}'
 $MUPIP trigger -triggerfile=trigger.trg >>& trigger.txt
 set rand='$'`$gtm_tst/com/genrandnumbers.csh 1 1 7`
 set fn=`$tst_awk "{print $rand}" fnlist.txt`
-set t1 = `date +"%b %e %H:%M:%S"`
+set t = `date +"%b %e %H:%M:%S"`
 $ydb_dist/mumps -run $fn^gtm8694>&output.txt
-set t2 = `date +"%b %e %H:%M:%S"`
 # Included to avoid the syntax error in the syslog created when calling getoper (which invokes an m function)
-sleep 1
-$gtm_tst/com/getoper.csh "$t1" "$t2" getoper.txt
+rm $ydb_dist/restrict.txt
+$gtm_tst/com/getoper.csh "$t" "" getoper.txt "" "RESTRICTSYNTAX"
 echo "# CHECKING THE SYSLOG"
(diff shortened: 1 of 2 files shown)


v63002/u_inref/gtm8694.csh
```diff
@@ -239,5 +240,26 @@ foreach restrict (`cat restrictlist.txt`)
...
+EOF
+chmod -w $ydb_dist/restrict.txt
+$gtm_tst/com/lsminusl.csh $ydb_dist/restrict.txt | $tst_awk '{print $1,$9}'
+$MUPIP trigger -triggerfile=trigger.trg >>& trigger.txt
+set rand='$'`$gtm_tst/com/genrandnumbers.csh 1 1 7`
+set fn=`$tst_awk "{print $rand}" fnlist.txt`
+set t1 = `date +"%b %e %H:%M:%S"`
+$ydb_dist/mumps -run $fn^gtm8694>&output.txt
+set t2 = `date +"%b %e %H:%M:%S"`
+# Included to avoid the syntax error in the syslog created when calling getoper (which invokes an m function)
```
nars1: I did not understand this comment. Seems out of place.

v63002/u_inref/gtm8694.csh
```diff
@@ -37,6 +37,7 @@ TRIGGER_MOD
 EOF
 # Not including CENABLE in this list because of logistics reasons
 echo "BREAK ZBREAK ZCMDLINE ZEDIT ZSYSTEM PIPE_OPEN DIRECT_MODE DSE TRIGGER_MOD">restrictlist.txt
+echo "breakfn zbreakfn zcmdlnefn zeditfn zsystemfn pipefn trigmodfn">fnlist.txt
```
nars1: Why was this change needed?

v63002/u_inref/gtm8694.csh
VinayYottaDB: There would be extra syntax errors in the syslog because getoper invokes m functions. To avoid this, i did a time interval (the sleep 1 prevents the getoper syslog messages from occurring in the same second as t2)

v63002/u_inref/gtm8694.csh
VinayYottaDB: This allows me to randomly select a function to execute for the last case

v63002/u_inref/gtm8694.csh
nars1: In that case, fix the restrict.txt syntax error before invoking getoper.csh. That way we keep the syntax error only for the mumps -run $fn^gtm8694 invocation.

v63002/u_inref/gtm8694.csh
nars1: Got it.

PR #365, opened 2018-08-06
New v63005/gtm8962 subtest (Tests GTM-8962 in V63005)

VinayYottaDB:

ZSHOW "I" outputs $ZPIN and $ZPOUT even if they are the same as $PRINCIPAL and no longer displays $ZPROCESS, as it was only meaningful on OpenVMS; also, $ZTNAME appears in proper alphabetic sequence. Previously if $ZPIN or $ZPOUT matched $PRINCIPAL, ZSHOW "I" omitted them, $ZPROCESS appeared as an empty string and $ZTNAME was not ordered appropriately. (GTM-8962)


v63005/outref/gtm8962.txt
```diff
@@ -0,0 +1,16 @@
+# Confirming $ZPIN=$PRINCIPAL, searching for $ZPIN in the output of ZSHOW "I"
+YDB>write:($ZPIN=$PRINCIPAL) "# $ZPIN = $PRINCIPAL"
+# $ZPIN = $PRINCIPAL
+$ZPIN="/dev/pts/3"
```
nars1: I want $PRINCIPAL grepped too. That is the only way we can confirm $ZPIN matches $PRINCIPAL. Same with $ZPOUT.

nars1: Also I suspect the 3 in /dev/pts/3 will not be the same every time we run this test. It depends on what other terminals are currently in use. So this might need to be TEST_AWKed out.

v63005/outref/gtm8962.txt
```diff
@@ -1,12 +1,13 @@
 # Confirming $ZPIN=$PRINCIPAL, searching for $ZPIN in the output of ZSHOW "I"
-YDB>write:($ZPIN=$PRINCIPAL) "# $ZPIN = $PRINCIPAL"
-# $ZPIN = $PRINCIPAL
-$ZPIN="/dev/pts/3"
+$ZPIN = $PRINCIPAL
+##TEST_AWK\$ZPIN="/dev/pts/[0-9]*"
 
 # Confirming $ZOUT=$PRINCIPAL, searching for $ZOUT in the output of ZSHOW "I"
```
nars1: ZOUT -> ZPOUT

PR #366, opened 2018-08-07
New v63005/gtm8961 subtest (Tests GTM-8961 in V63005)

VinayYottaDB:

The path reported by ZSHOW "D", $KEY, and $ZSOCKET("","LOCALADDRESS",) for LOCAL sockets passed via JOB or WRITE /PASS is correct. Previously on AIX 7.2, there could have been extra bytes at the end. This issue was only observed in the GT.M development environment, and was never reported by a user.[AIX 7.2] (GTM-8961)


v63005/inref/gtm8961.m
```diff
@@ -0,0 +1,45 @@
...
+;	of its copyright holder(s), and is made available	;
+;	under a license.  If you do not know the terms of	;
+;	the license, please stop and do not read further.	;
+;								;
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;
+
+writeslashpass
+	set ^stop=0
+        set sock="socket"
```
nars1: Lines 16 and 17 seem to have alignment off (relative to lines 15 and 18).

PR #369, opened 2018-08-08
New v63005/gtm8959 subtest (Tests GTM-8959 in V63005)

VinayYottaDB:

ZGOTO 0 in a call-in returns to the invoking program. Previously on x86-64 Linux systems with glibc 2.24 or later, this generated a segmentation violation (SIG-11). (GTM-8959)


v63005/u_inref/gtm8959.csh
```diff
@@ -0,0 +1,49 @@
...
+$gt_ld_linker $gt_ld_option_output gtm8959ci $gt_ld_options_common gtm8959ci.o $gt_ld_sysrtns $ci_ldpath$gtm_dist -L$gtm_dist -lgtmshr $gt_ld_syslibs >&! link2.map
+if ($status) then
+	echo "# Creation of Executable failed"
+endif
+echo "# Running a c program that will execute a call in with a ZGOTO 0, this would previously produce a SIG-11 error"
+./gtm8959ci
+if ($status) then
+	echo "# Error encountered while running program"
+else
+	echo "# No Sig-11 error"
```
nars1: echo "# Status 0 returned from gtm8959ci" instead of saying  No Sig-11 error.

v63005/u_inref/gtm8959.csh
```diff
@@ -0,0 +1,49 @@
...
+
+cat > gtm8959ci.c << EOF
+#include <string.h>
+#include <limits.h>
+#include "gtm_stdio.h"
+#include "gtmxc_types.h"
+int main()
+{
+	int status = gtm_ci("gtm8959");
+	return 0;
```
nars1: Can we add a printf line here to say control does return back to the C program.

PR #374, opened 2018-08-09
New v63005/gtm5574 subtest (Tests GTM-5574 in V63005)

VinayYottaDB:

The %DH,%HD,%OH,%HO,%OD, and %DO routines now handle conversions with numbers up to GT.M's maximum string length in size. Additionally, %DH and %DO routines now handle the conversion of negative decimal numbers properly, even when the specified length is not long enough to represent that fully converted number. Previously, the %DH,%HD,%OH,%HO,%OD, and %DO routines could only deal with numbers up to the equivalent of the maximum 18-digit decimal number. Attempting to convert a number larger than this would cause roundoff in the final result, which still occurs if they are used for arithmetic. Moreover, specifying a length for the %DH and %DO routines that was not long enough to hold the fully converted result would cause the routines to produce incorrect values. The workaround for both these issues was to avoid using these percent routines with numbers greater than the maximum 18-digit decimal number and to always specify an appropriate length for the %DH and %DO routines. (GTM-5574)


v63005/outref/gtm5574.txt
```diff
@@ -0,0 +1,64 @@
...
+%HD-----Correct Conversion
+-------------------------------------------
+# Conversions using a random 10 digit negative number
+%DH-----Correct Conversion
+%HO-----Correct Conversion
+%OD-----Correct Conversion
+%DO-----Correct Conversion
+%OH-----Correct Conversion
+%HD-----Correct Conversion
+-------------------------------------------
```
nars1: Can we also test the bolded part in the release note.

Additionally, %DH and %DO routines now handle the conversion of negative decimal numbers properly, **even when the specified length is not long enough to represent that fully converted number**.

v63005/u_inref/gtm5574.csh
```diff
@@ -0,0 +1,31 @@
...
+#	of its copyright holder(s), and is made available	#
+#	under a license.  If you do not know the terms of	#
+#	the license, please stop and do not read further.	#
+#								#
+#################################################################
+#
+#
+#
+
+# Set to avoid wrapping from bc
```
nars1: Replace with "# set to avoid line wrapping at 68 columns or so which breaks long numbers into multiple lines otherwise"

v63005/inref/gtm5574.m
```diff
@@ -26,13 +26,32 @@
...
 	close o
-	set fourgig=4294967296
+	if $ZCMDLINE<0 do neg
+	else  do pos
+	quit
+
+neg
+	set %DH=n
+	set %DO=n
+	set %DL=1
```
nars1: Can you set %DL to a random value between 1 and 16 and randomly unset it too. All of them should continue to pass the existing tests. This way we test more cases.

v63005/inref/gtm5574.m
```diff
@@ -33,7 +33,8 @@
 neg
 	set %DH=n
 	set %DO=n
-	set %DL=1
+	kill %DL
+	set:($random(1)) %DL=1+$random(16)
```
nars1: $random(1) -> $random(2)
$random(1) always returns 0 so the set command will never execute

PR #341, opened 2018-07-25
New r124/ydb312_gtm8182g subtest (tests YottaDB/YottaDB#312 in r1.24)

JakeYottaDB:
Sets up and tests the following test case 

* Add test case where DB1 is in GLD1 and DB2 is in GLD2 and DB1/DB2 are in GLD3. And source server is started with GLD3.

  And all of GLD1/GLD2/GLD3 use the same mumps.repl file. Now a process that updates through GLD1 and GLD2 should

  attach to the journal pool only once. And the ftok semaphore counter should be bumped only once, not twice.

PR #375, opened 2018-08-14
[#341] New r124/ydb341 subtest (tests YottaDB/YottaDB#341)

nars1:


estess (commented):
Approved.

But what's with the strange indenting that indents some (mostly SETs but not of them) and not others by 2 bytes?

nars1:
Sorry. It is tabs vs spaces. Will fix that next time I am in that module.

